PR #5, opened 2017-11-17
Fixing bug in SingleListGrader with ordered input

jolyonb:
I found a bug :-(

If ordered was True on SingleListGrader, then it wasn't handling the wrong number of inputs correctly.

This squashes the bug.


jolyonb:
I'm merging this, because I need that zipfile for the website. Note that I've also been pushing directly to master for writing the website... it's almost done!

jolyonb:
(You should feel free to check over the code adjustments though!)

Issue #9, opened 2017-11-30
FormulaGrader throws an error when passed an empty formula


jolyonb:
This gives an error rather than just grading incorrect.

````
from mitxgraders import *
grader = FormulaGrader(
    answers="1"
)
print grader(None, "")
````

ChristopherChudzicki:
Sorry I didn’t comment on this when you first posted, I just saw it referenced in #13.

I’mThinkingAboutThisNow, andIsAnErrorOnEmptyStreamsSubmission ofG sorry I didn’t comment on this when you first posted, I just saw it  referenced in #13.

 I’m thing about this now, and is this actually undesirable?  Might it not be better to universally raise an error on empty string submissions (For all graders, with useful error message and perhaps a configuration key override). Also isn’t that what edx does by default now?  I believe the rationale is to not use up student attempts if they accidentally submit an empty string.

jolyonb:
When you have a large number of attempts and a two-part-in-one question, it's common to submit the first part of the question, leaving the second blank. This is the situation in which I noticed this error. I think the only way to deal with this is to mark an empty formula incorrect. Furthermore, edX won't let you submit an empty box if that's the only part to the question.

ChristopherChudzicki:
Oh perfect— I had forgotten that the submit button was disabled rather than just raising an error. This sounds good then.  I can review this and other PRs tonight after dinner or tomorrow.

ChristopherChudzicki:
Resolved by #13

PR #12, opened 2017-12-15
Fixing bug when empty string submitted to FormulaGrader

jolyonb:
An empty string now grades as wrong. This addresses issue #9.


jolyonb:
Found an issue in our test suite... fixing!

Issue #10, opened 2017-11-30
Dependent Sampling


jolyonb:
Another idea that may be useful is dependent sampling. Eg, sample x, y, z, then set r equal to sqrt(x^2 + y^2 + z^2) instead of sampling it separately. We'd have to be careful about detecting circular dependencies.

ChristopherChudzicki:
Resolved by #15

PR #15, opened 2017-12-16
Dependent sampling

jolyonb:
This addresses issue #10 and creates a sampler that depends on other variables. Documentation is updated, including the example course (which is already updated). Tests also included.

I ran into a bunch of circular reference issues between mathfunc.py and sampling.py, so I ended up moving some helper functions into sampling.py.

I'm pretty happy with both the API and implementation of DependentSampler. I'm less happy about the implementation of gen_symbols_samples. Very happy to take comments there.


mitxgraders/sampling.py
```diff
@@ -331,3 +342,243 @@ class SpecificFunctions(FunctionSamplingSet):  # pylint: disable=too-few-public-
...
+    ]
+
+    # Generate the samples
+    sample_list = []
+    for _ in range(samples):
+        # Generate independent samples
+        sample_dict = {symbol: sample_from[symbol].gen_sample() for symbol in independent}
+
+        # Generate dependent samples, following chains as necessary
+        dependent = [
```
ChristopherChudzicki: I initially did not understand why this was inside the for loop (it gets mutated below). So so I moved it outside the for loop. None of the tests broke, because all tests use `samples=1`. So I would suggest changing one of the tests to having `samples=2`.

ChristopherChudzicki: Re API, I am also happy. Re implementation, the overall approach seems fine—multiple passes through the list, evaluating symbols if possible. One could also imagine building some sort of tree structure that reflects the variable dependencies, but this works fine. 

While trying to understand the code, I did refactored a few things that resulted in a somewhat less nesting. It also stored the dependent variables in a dictionary and deleted their keys after evaluation, so there's no need to mark evaluated symbols with empty strings. Here's my version:

```python
def is_subset(iterable, iterable_superset):
    for item in iterable:
        if item not in iterable_superset:
            return False

    return True
```
Then:
```python
# Generate dependent samples, following chains as necessary
unevaluated_dependents = {
    symbol: sample_from[symbol].config['depends'] for symbol in symbols
    if isinstance(sample_from[symbol], DependentSampler)
}
while unevaluated_dependents:
    progress_made = False
    for symbol, dependencies in unevaluated_dependents.items():
        if is_subset(dependencies, sample_dict):
            sample_dict[symbol] = sample_from[symbol].compute_sample(sample_dict)
            del unevaluated_dependents[symbol]
            progress_made = True
        else:
            continue

    if not progress_made:
        bad_symbols = ", ".join(sorted(unevaluated_dependents.keys()))
        raise ConfigError("Circularly dependent DependentSamplers detected: " + bad_symbols)
```

jolyonb:
Much cleaner. Thanks for the refactoring. I've basically copied your code in directly, and updated the tests so that it is tested for multiple samples.

PR #16, opened 2017-12-16
use test_zip.py's path to locate python_lib.zip

ChristopherChudzicki:
What the title says.


jolyonb:
Aw man, that's just embarrassing that I left that in there! I've added the zip test back into the travis test suite.

PR #19, opened 2017-12-17
Adding tests to increase coverage

jolyonb:
In the spirit of procrastination, I've added some tests to increase coverage of the library. Lo and behold, as I did so, I found a handful of bugs! I also took the opportunity to do a little linting.

We now have 100% coverage! (minus one line...)

### DIFF
--- a/mitxgraders/helpers/calc.py
+++ b/mitxgraders/helpers/calc.py
@@ -102,8 +102,8 @@ def handle_np_floating_errors(err, flag):
         raise OverflowError
     elif 'value' in err:
         raise ValueError
-    else:
-        raise Exception(err) # should not happen
+    else:  # pragma: no cover
+        raise Exception(err)
 
 np.seterrcall(handle_np_floating_errors)
 np.seterr(divide='call', over='call', invalid='call')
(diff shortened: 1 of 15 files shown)


jolyonb:
So, it turns out there's a single line that we can't test in baseclasses.py, because ListGrader cannot presently return an overall message. There's a few other places where I added "pragma: no cover", as the error checking code in those places should never be triggered, or it requires zip file loading. Otherwise, I fixed some bugs, removed some redundant code, and did a bit of linting. Should be ready to merge!

mitxgraders/helpers/calc.py
```diff
@@ -478,7 +478,7 @@ def parse_algebra(self):
         expr << sum_term  # pylint: disable=pointless-statement
         self.tree = (expr + stringEnd).parseString(self.math_expr)[0]
 
-    def reduce_tree(self, handle_actions, terminal_converter=None):
```
jolyonb: We're not using this, and I don't think we ever will - it's a holdover from the edX code (that wasn't using it either!)

mitxgraders/helpers/validatorfuncs.py
```diff
@@ -94,7 +94,7 @@ def func(config_input):
             config_input = (config_input,)
         # Apply the schema
         if validator:
-            schema = Schema(All((given_type,), Length(min=1), [validator]))
+            schema = Schema(All((given_type,), Length(min=1), (validator, )))
```
jolyonb: This was a bug. We weren't using this line though.

mitxgraders/listgrader.py
```diff
@@ -324,10 +324,6 @@ def validate_grouping(self):
                   "or a list of subgraders"
             raise ConfigError(msg)
 
-        # Unordered, we must have a single ListGrader subgrader
```
jolyonb: Redundant error checking; this is tested for previously, as it applies regardless of whether or not grouping is used.

Issue #11, opened 2017-12-06
Request for variable names with superscripts


jolyonb:
See title! Not obvious how you'd implement this...

ChristopherChudzicki:
Curious: What's the use-case? Tensor notation? Superscript notation for derivatives? Do people just want `f^2`, or will they way `f^(2,1)` as a variable also?

First: Certainly if `x^2` is a variable, then `x` should not be a variable also, so we'd have to add some validation for the schema. (I suppose that if `f^2` is a function, `f` could also be a function).

Here are two possible thoughts for implementation:

-  Pre-process the user-submission before passing it to the parser. That is, replace all instances of the declared variable `x^2` with a valid (unused) variable like (`x1`). This might complicate sampling and error messages. This seems a bit messy.
- Alter the parser based on the grader configuration. For example, if `x^2` and `y^2` are declared as variables, alter the parser to use

```python
superscript_vars = ['x^2', 'y^2'] # would be passed as an argument
inner_varname = Combine(
    Or(superscript_vars) |
    (
     Word(alphas + "_", alphanums + "_") +
     ZeroOrMore("'")
    )
)
```

This correctly recognizes `a^2` as "a squared", but treats `x^2` and `y^2` as single variables:

```XML
<sum>
  <product>
    <parallel>
      <power>
        <atom>
          <variable>
            <ITEM>a</ITEM>
          </variable>
        </atom>
        <ITEM>^</ITEM>
        <atom>
          <number>
            <ITEM>2</ITEM>
          </number>
        </atom>
      </power>
    </parallel>
  </product>
  <ITEM>+</ITEM>
  <product>
    <parallel>
      <power>
        <atom>
          <variable>
            <ITEM>y^2</ITEM>
          </variable>
        </atom>
      </power>
    </parallel>
  </product>
  <ITEM>+</ITEM>
  <product>
    <parallel>
      <power>
        <atom>
          <variable>
            <ITEM>x^2</ITEM>
          </variable>
        </atom>
      </power>
    </parallel>
  </product>
</sum>
```

Caveat about the second approach: The order of parsing terms in inner_varname seems to be important—superscript_vars needs to come before the normal variables—and I do not completely understand why that is yet.

jolyonb:
Simona wanted to have tensor notation. I'm not sure if she wanted abstract index notation or just specific component notation. I was thinking something like a^{ijk} might be possible, using the curly braces to specify that indices were coming.

[...]

PR #24, opened 2018-01-03
Enabling tensor variable names

jolyonb:
Enables tensor variable names:

* T_{ijk}
* T^{ijk}
* T_{ijk}^{123}

Version number, zip file, documentation and example course are all updated.

Addresses Issue #11.


jolyonb:
@ChristopherChudzicki While I was in here, I was reminded that you wanted to rewrite the function evaluator at some point, and that there was a technical issue with the present one? (apart from being woefully inefficient...)

ChristopherChudzicki:
Re 
> @ChristopherChudzicki While I was in here, I was reminded that you wanted to rewrite the function evaluator at some point, and that there was a technical issue with the present one? (apart from being woefully inefficient...)

Since the first version of FormulaGrader I have wanted to be able to parse expressions like `[1,2,3]` or `[[1,2], [3,4]]` as vectors/matrices (for reasons I can elaborate on somewhere else).

mitxgraders/helpers/calc.py
```diff
@@ -446,10 +447,18 @@ def parse_algebra(self):
...
+        inner_varname = Combine(Word(alphas, alphanums + "_") + ZeroOrMore("'"))
+        # Alternative variable name in tensor format
+        # Tensor name must be alphas only, indices may be alphanumeric
+        # e.g., U_{ijk}^{123}
+        indices = Literal("{") + Word(alphanums) + Literal("}")
+        tensor_lower = Combine(Word(alphas) + Literal("_") + indices)
+        tensor_upper = Combine(Word(alphas) + Literal("^") + indices)
+        tensor_mixed = Combine(Word(alphas) + Literal("_") + indices + Literal("^") + indices)
+        # Try for tensor mixed first, then upper, then lower, then generic variable
+        varname = Group(tensor_mixed | tensor_upper | tensor_lower | inner_varname)("variable")
```
ChristopherChudzicki: Do you want to allow primes after a tensor? I think this might be good. Also, if primes are allowed, I think this could be simplified to

```
inner_varname = Combine(Word(alphas, alphanums + "_") + ZeroOrMore("'"))
# Alternative variable name in tensor format
# Tensor name must be alphas only, indices may be alphanumeric
# e.g., U_{ijk}^{123}
upper_indices = Literal("^{") + Word(alphanums) + Literal("}")
lower_indices = Literal("_{") + Word(alphanums) + Literal("}")
tensor = Combine( Word(alphanums) + Optional(lower_indices) + Optional(upper_indices) + ZeroOrMore("'") )
# Try for tensor mixed first, then upper, then lower, then generic variable
varname = Group(tensor | inner_varname)("variable")
```

(In fact, in this case, I believe `tensor` will almost always match. The only case where I believe `inner_varname` would match is variables with multiple  underscores: `a_1_2_3` or `b____4`. I don't see a reason to even allow such variables, except that edX does.)

ChristopherChudzicki: Also, in an ideal world, I think we would allow `T_i^{123}` and `T_{i}^{123}` **and interpret them the same way**.

(To be clear, if we can't interpret them the same, then I think disallowing `T_i^{123}` is the best option.)

This could be a future enhancement if it is desired.

tests/test_base.py
```diff
@@ -306,3 +306,10 @@ def test_math():
     assert -arcsech(sech(x)) == approx(x)
     assert arccsch(csch(x)) == approx(x)
     assert arccoth(coth(x)) == approx(x)
+
+def test_tensors():
+    """Test tensor variable names in calc.py"""
+    assert evaluator("U^{ijk}", {"U^{ijk}":2}, {}, {})[0] == 2
+    assert evaluator("U_{ijk}/2", {"U_{ijk}":2}, {}, {})[0] == 1
+    assert evaluator("U_{ijk}^{123}", {"U_{ijk}^{123}":2}, {}, {})[0] == 2
+    assert evaluator("U_{ijk}^2", {"U_{ijk}":2}, {}, {})[0] == 4
```
ChristopherChudzicki: Maybe add a test to show that powers still work?

```python
 assert evaluator("U_{ijk}^{123}^2", {"U_{ijk}^{123}":2}, {}, {})[0] == 4
```

jolyonb:
Was there also a technical error in the parser? Something about orders of operations with nested powers? Happy to discuss in an issue. Looking at your other comments now.

mitxgraders/helpers/calc.py
jolyonb: I thought about primed tensors and decided not to, but since you've provided the code... One change I'm making is in the tensor= line, changing Word(alphanums) to Word(alphas, alphanums).

jolyonb: I think I want to keep braces to indicate indices for tensors. If we ever make a special sampler for tensors (eg, specifying traces, symmetries, etc), we're going to want tensors to be very specific. Equally so if we ever do abstract index notation for relativity.

tests/test_base.py
jolyonb: The fourth line already does this!

jolyonb:
Changed code to add in primes per your suggestion. Added test. Remade zip file. Rebased. I think this is ready to merge!

jolyonb:
Ok, tests are failing. That's embarrassing.

jolyonb:
Ok, should be good to go now! (typo in the new test...)

tests/test_base.py
ChristopherChudzicki: OK, I wasn't sure if you wanted one that used both upper indices and powers. (4th doesn't use upper indices now).

ChristopherChudzicki:
Re

> Was there also a technical error in the parser? Something about orders of operations with nested powers? Happy to discuss in an issue. Looking at your other comments now.

Oh, yes. I can make an issue for it later. The issue is with signs in the exponents: `2^-2^2` displays as `2^-(2^2)` but parses as`2^(-2)^2`. I don't think it's a very big deal, because it only happens for numbers, not for variables: `e^-x^2` is not a valid expression, so it does not get confused with `e^(-x)^2`. 

Again, I'll write that more coherently later.

jolyonb:
I've added some extra tests to show that powers still work.

Issue #25, opened 2018-01-31
Parser doesn't work properly


jolyonb:
The tensor parser broke the old parsing. The following no longer works:

````python
import mitxgraders
evalf = mitxgraders.helpers.calc.evaluator
print(evalf("epsilon_max", {"epsilon_max":2}, {}, {}))
````

The error is raised by pyparsing; the old variable names with underscores aren't being accepted.

**Comment by jolyonb** [comment]
So, it turns out that the tensor parser is eating "epsilon", so the old parser never gets a chance. I think we need to revert to my version of the code...

**Comment by ChristopherChudzicki** [comment]
Ugh, you're right. I'm surprised we didn't have a subscript in a FormulaGrader test ...

**Comment by ChristopherChudzicki** [comment]
Here's another option: `x ^ y` ("Or") matches x or y, whichever is longer. 

```
upper_indices = Literal("^{") + Word(alphanums) + Literal("}")
lower_indices = Literal("_{") + Word(alphanums) + Literal("}")
indices = lower_indices ^ ( Optional(lower_indices) + upper_indices )
tensor = Combine(Word(alphas, alphanums) + indices + ZeroOrMore("'"))
```

jolyonb:
I'm tempted to keep this one simpler. pyparsing is complicated enough, and I'm not sure that the added sophistication is worth it, compared to something that is much more straightforward to interpret.

PR #27, opened 2018-02-20
Reorganize variable name parsing [WIP, do not merge]

ChristopherChudzicki:
This is a rewrite of the variable name parsing to address #25 

This is a bit more flexible than @jolyonb's original tensor specification. Anything of the form

```
front + subscripts + lower_indices + upper_indices + tail
```
is allowed, where:
- `front` (required): Alphabetical followed by alphanumeric, e.g., `Xab`
- `subscripts` (optional): Alphanumeric and underscores, e.g., `_cd___2_3`
- `lower_indices ` (optional): Like `_{12xy3}`
- `upper_indices ` (optional): Like `^{4abc}`
- `tail ` (optional): Any number of primes.

So for example, `Xab _cd___2_3_{12xy3}^{4abc}` is an allowed variable name.

Using the language above, edX original is `front+start`

### DIFF
--- a/mitxgraders/helpers/calc.py
+++ b/mitxgraders/helpers/calc.py
@@ -23,6 +23,7 @@ from pyparsing import (
     ParseResults,
     Suppress,
     Word,
+    FollowedBy,
     ZeroOrMore,
     alphanums,
     alphas,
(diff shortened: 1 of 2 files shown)


jolyonb:
That looks reasonable to me. I have no idea how edX will display something like abc_stuff_{ijk}, but we can certainly allow it in the parsing. Thanks for getting to this - it's been sitting on my todo list for a while...

ChristopherChudzicki:
Glad it seems reasonable. I think that using `abc_stuff_{ijk}` is probably not a good idea, but allowing it in the parser seems natural and seems to make the logic simpler: now tensor notation is a strictly "extra stuff" added to end of original edX variable name.

I'm currently happy with the implementation, but I'd like to write some more tests before merging.

mitxgraders/helpers/calc.py
```diff
@@ -447,20 +448,34 @@ def parse_algebra(self):
...
-        # compatability
-        varname = Group(tensor | inner_varname)("variable")
+        upper_indices = Literal("^{") + Word(alphanums) + Literal("}")
+        tail = ZeroOrMore("'")
+        inner_varname = Combine(
+            front +
+            Optional(subscripts) +
+            Optional(lower_indices) +
+            Optional(upper_indices) +
+            Optional(tail)
```
jolyonb: I think this doesn't need to be optional on the tail?

ChristopherChudzicki: True. It's effectively optional already by `ZeroOrMore`. I'll switch it to a comment.

jolyonb:
Indeed. Please include tests like `a2bc_d3e__4fg_{ijk}`!

ChristopherChudzicki:
@jolyonb 

> Indeed. Please include tests like a2bc_d3e__4fg_{ijk}!

Maybe you realized that wouldn't work :) 

I had trouble implementing normal subscripts and tensor lower_indices simultaneously for the same variable. Since I don't think they would display well anyway, I don't want to worry about it.

I've updated the PR description. The parsing is now, I believe, what you originally envisioned and is fairly clean. So, please review and merge if no issues.

PR #36, opened 2018-06-29
refactor forbidden string check and move after correctness check

### LINKED ISSUE #33: Strip whitespace from forbidden strings
In FormulaGrader, we strip whitespace from student expression before checking for forbidden strings. We should also strip whitespace from the forbidden strings themselves.

For example,

```python
# Simplify (a + b)^2 - (a - b)^2
FormulaGrader(
    answers='4*a*b',
    variables=['a', 'b']
    forbidden_strings=['a + b', 'a - b'],
)
```
would currently accept `(a+b)^2 - (a    -    b)^2` as an answer.

Additionally, let's refactor the validation into a static method.

(I refactored usage of required functions into a static method as part of #32, but am leaving this separate since there would be an actual behavior change.)

ChristopherChudzicki:
This resolves #33 and moves the forbidden string check after the correctness check, as requested by Stepan at Harvard:

> ... extremely useful, especially the "forbidden string" feature for problems where we ask users to simplify an expression like "a^2 * a^b" to a single exponential by forbidding "*".
>
> However, we have run into an issue that we want to get your thoughts on. The package currently checks first for forbidden strings, and then for correct evaluation. For the problem above, a response of "a^(2*b)" would fail both checks, but since the forbidden string test happens first, the user gets a yellow-light message like "Please simplify completely." We could change the yellow-light message to something more generic, but we think a more natural solution would be to perform the evaluation check first (breaking to a red-light "Incorrect" message if it fails), and the forbidden string check second. We wanted to hear your thoughts on this scheme before trying to implement it, and see if you had run into similar issues.

### DIFF
--- a/mitxgraders/formulagrader.py
+++ b/mitxgraders/formulagrader.py
@@ -25,6 +25,10 @@ __all__ = [
     "CalcError"
 ]
 
+def casify(thestring, case_sensitive):
+    """Converts a string to lowercase if not case_sensitive"""
+    return thestring if case_sensitive else thestring.lower()
+
 class FormulaGrader(ItemGrader):
     """
     Grades mathematical expressions, like edX FormulaResponse.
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki:
@jolyonb Regarding the idea of moving whitelist/blacklist check after the check function, I think that actually is doable if we wanted to.

- Currently whitelist/blacklist is used to construct `self.functions`, which is passed to evaluator.
 - Couldn't we instead just not restrict the functions passed to evaluator, then check after the fact that no blacklist or only whitelist functions were used? Evaluator returns a list of a list `used_functions`, so I think this would be possible (and possibly simpler than what we do now)... But maybe I'm missing something

jolyonb:
True, we can definitely refactor the black/whitelist functions in that manner. It would bring the treatment in line with the forbidden strings too. Feel free to give it a go in another PR?

PR #49, opened 2018-06-30
Introduces sampling for numbered variables

jolyonb:
This introduces the concept of a numbered variable in FormulaGrader. A separate key, `numbered_vars` is used to set up numbered variables, which leverage the existing `sample_from` infrastructure. Here is an example of its use:

````python
grader = FormulaGrader(
    answers='a_{0} + a_{1} + a_{-1}',
    numbered_vars=['a'],
    sample_from={'a': [0,5]}
)
````

Any integer entered as an index will be accepted by the grader; there is presently no way of restricting it. I don't think this is actually necessary; if a student enters a_{-1} when only positive indices have meaning, they will be (correctly) marked wrong.

There is a little duplication happening in constructing and checking the list of used variables between FormulaGrader and `calc.py`. That functionality needs to remain in `calc.py` so it can be used outside of FormulaGrader, but it also needs to be used to construct the numbered variables. I think it's ok.

Note that there is a very small expansion of acceptable variable names in the form of the optional inclusion of a minus sign in tensor notation.

Documentation has been updated, and an example added to the sample course.

### CODE REVIEW

File: mitxgraders/formulagrader.py [hunk 1]
```diff
@@ -437,11 +446,58 @@ def check_response(self, answer, student_input):
...
+        variable_list = self.config['variables'][:]
+        sample_from_dict = self.config['sample_from'].copy()
+
+        # Find all unassigned variables
+        cs = self.config['case_sensitive']
+        bad_vars = set(var for var in vars_used if casify(var, cs) not in variable_list)
+
+        # Check to see if any unassigned variables are numbered_vars
+        for var in bad_vars:
+            csvar = casify(var, cs)
```
ChristopherChudzicki: Looking through this now. I got stuck on the next few lines for quite a while :). 

I think it's worth abstracting the next several lines into a separate function for (1) clarity and (2) testing that isolated bit of logic. Maybe something like:

```python
for var in bad_vars:
    if numbered_variables.match(var):
        # do stuff
```

Also, I think it will be more understandable to just construct a single regexp for matching the whole numbered_variable. With the help of SO, I made this function that I think does what we need (sans casification, but that either shouldn't be hard or we'll drop it.)

```python
import re

def numbered_vars_regexp(variable_fronts):
    """
    Creates a regexp to match numbered variables.

    Arguments:
        variable_fronts ([str]): a list of variable fronts

    Usage
    =====

    Matches numbered variables:
    >>> regexp = numbered_vars_regexp(['b', 'c', 'cat'])
    >>> regexp.match('b_{12}').groups()
    ('b_{12}',)
    >>> regexp.match('b_{-3}').groups()
    ('b_{-3}',)
    >>> regexp.match('b_{0}').groups()
    ('b_{0}',)

    Other variables match, too:
    >>> regexp.match('cat_{17}').groups()
    ('cat_{17}',)

    Stuff that shouldn't match does not match:
    >>> regexp.match('b') == None
    True
    >>> regexp.match('b_{05}') == None
    True
    >>> regexp.match('b_{-05}') == None
    True
    """
    # Adapted from https://stackoverflow.com/a/33411504/2747370

    # I'm not sure that the sorting is really necessary, since we match underscore next
    longest_first = sorted(variable_fronts, key=len, reverse=False)
    base_regexp = (r"^((?:{longest_first})" # match base
                   "_{{" # match _{, double curly b/c of format string
                   "(?:[-]?[1-9]\d*|0)" # match number pattern
                   "}})$") # match closing }
    return re.compile(base_regexp.format(
        longest_first = '|'.join(map(re.escape, longest_first))
    ))

```

jolyonb: Yeah, I realized I could do the whole thing with regex, but knew how to strip off some of the simpler bits without getting into the mud of regex ;-) I'll go and make that change now.

ChristopherChudzicki:
Cool, I thought this was going to be harder! 

I think it might be worth testing a few cases (positive/negative/zero), but if you decide to use the `numbered_vars_regexp()` function above, the doctests could take care of that.

ChristopherChudzicki:
Oh, and I suspect you were just waiting to finish the review, but changelog and zip!

(At some point, it might be worthwhile merging things like this into a `next` branch instead of master, so we can have fewer minor releases.)

mitxgraders/formulagrader.py
ChristopherChudzicki: > without getting into the mud of regex ;-) 

Yes, I was a bit wary of wading into that myself. I heard a story about a guy who had a problem, and tried to solve it with a regex, and then he had two problems. (That's supposed to be a joke.) But I think it's worth it here, especially for making an easier-to-test bit of logic.

jolyonb:
Ok, see PR #50. After a decision is made on that, I'll rebase and update this appropriately. I think if #50 is accepted, this should be v1.1.

Issue #26, opened 2018-02-01
Issues with variable name collision when case insensitive


jolyonb:
When a user defines a variable `E` and specifies case insensitive, it's random as to whether `e = E` or `e = 2.78...`.

jolyonb:
I believe that our thinking is that user-defined constants and variables override default constants. If `E` is defined in case insensitive, I believe this should override the default constant `e`.

I think a `ConfigError` should be raised if there is a collision between user-defined variables/constants. This check should be done for case sensitive and insensitive.

ChristopherChudzicki:
I think that makes senses. If we put matrix grading into FormulaGrader, I'm not sure what to do about `i` and `I` (identity) and `case_sensitive=False`.  That said, I've been thinking more that we should have a separate MatrixFormulaGrader with `I` predefined, leave `I` undefined in regular FormulaGrader, and disallow case-sensitive in MatrixFormulaGrader.

jolyonb:
That makes sense. However, I'll propose an alternative: remove case insensitive entry altogether. We don't know anybody who uses it, it causes collisions like this, it complicates the code, and it has ambiguities that I think we'd just rather avoid. What do you think?

ChristopherChudzicki:
> However, I'll propose an alternative: remove case insensitive entry altogether.
I think that's reasonable, although it would make FormulaGrader no longer a superset of <Formularesponse />. Worth thinking about more, maybe asking Harvard people and Jen (though I'm about 100% sure she doesn't use it).

I have other reasons for suggesting a separate MatrixFormulaGrader class which I'll expand on next week. (One though I have for implementing transpose is by means of defining a `T` variable, which might not be something we want present in FormulaGrader by default.)

jolyonb:
Ok, I can definitely understand wanting a transpose operator to be only available to matrix problems.

Anyway, back on task here. I made a PR to remove the case-insensitive option, which I think is something we should move ahead with. That mostly fixes the issue here, but leaves a residual issue: we should still check to make sure that we don't have variable name collisions (although overwriting the default constants is fine).

PR #50, opened 2018-07-01
Removing case-insensitive input for FormulaGrader, IntegralGrader

jolyonb:
Partially addresses #26. Documentation has been updated appropriately. Also fixes some issues in tests.

### DIFF
--- a/mitxgraders/formulagrader.py
+++ b/mitxgraders/formulagrader.py
@@ -27,10 +27,6 @@ __all__ = [
     "CalcError"
 ]
 
-def casify(thestring, case_sensitive):
-    """Converts a string to lowercase if not case_sensitive"""
-    return thestring if case_sensitive else thestring.lower()
-
 # Some of these validators are useful to other classes, e.g., IntegralGrader
 def validate_blacklist_whitelist_config(blacklist, whitelist):
     """Validates the whitelist/blacklist configuration.
@@ -56,7 +52,7 @@ def validate_blacklist_whitelist_config(blacklist, whitelist):
         if func not in DEFAULT_FUNCTIONS:
             raise ConfigError("Unknown function in whitelist: {func}".format(func=func))
 
-def validate_forbidden_strings_not_used(expr, forbidden_strings, forbidden_msg, case_sensitive):
+def validate_forbidden_strings_not_used(expr, forbidden_strings, forbidden_msg):
     """
     Ignoring whitespace, checking that expr does not contain any forbidden_strings.
     Usage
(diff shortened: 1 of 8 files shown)


mitxgraders/formulagrader.py
```diff
@@ -65,54 +61,50 @@ def validate_forbidden_strings_not_used(expr, forbidden_strings, forbidden_msg,
...
-    ... set(['f', 'g', 'sin', 'cos']),
-    ... True)
+    ... set(['f', 'g', 'sin', 'cos'])
+    ... )
     Traceback (most recent call last):
     InvalidInput: Invalid Input: function(s) 'h', 'Sin' not permitted in answer
     """
-    casified_permitted = {casify(f, case_sensitive) for f in permitted_functions}
+    casified_permitted = {f for f in permitted_functions}
     used_not_permitted = [f for f in used_funcs
```
ChristopherChudzicki: Let's get rid of `casified_permitted` entirely since it's not casified anymore, and just use

```
used_not_permitted = [f for f in used_funcs
                          if f not in permitted_functions]
```

mitxgraders/formulagrader.py
```diff
@@ -27,10 +27,6 @@
     "CalcError"
 ]
 
-def casify(thestring, case_sensitive):
-    """Converts a string to lowercase if not case_sensitive"""
-    return thestring if case_sensitive else thestring.lower()
-
 # Some of these validators are useful to other classes, e.g., IntegralGrader
 def validate_blacklist_whitelist_config(blacklist, whitelist):
```
ChristopherChudzicki: Line 38 has a comment about casifying. Delete it?

(Apparently I can't comment on line 38 itself since its not close enough to a change ... weird.)

mitxgraders/helpers/calc.py
```diff
@@ -542,38 +529,43 @@ def check_variables(self, valid_variables, valid_functions):
...
 
-        # Test if casify(X) is valid, but return the actual bad input (i.e. X)
         bad_vars = set(var for var in self.variables_used
-                       if casify(var) not in valid_variables)
+                       if var not in valid_variables)
         if bad_vars:
             message = "Invalid Input: {} not permitted in answer as a variable"
             varnames = ", ".join(sorted(bad_vars))
+
+            # Check to see if there is a different case version of the variable
```
ChristopherChudzicki: Nice. This is just generally helpful, plus it makes me feel better about removing case_sensitive.

mitxgraders/formulagrader.py
jolyonb: Yup, missed that entirely. I was searching for "case" and "casify" ;-)

mitxgraders/formulagrader.py
jolyonb: Consider it gone!

PR #52, opened 2018-07-01
bugfix: formulagrader debug log

ChristopherChudzicki:
FormulaGrader debug log used to have a hard-coded key name...Oops.


tests/test_formulagrader.py
```diff
@@ -492,57 +492,61 @@ def test_fg_debug_log():
...
+            'square': lambda x: x**2
+            },
         samples=2,
         debug=True
     )
     result = grader(None, 'z + x*x + f(y)')
 
     message = (
-        "<pre>MITx Grading Library Version {version}<br/>\n"
+        "<pre>MITx Grading Library Version 1.1<br/>\n"
```
jolyonb: Surely you want this to be read from the version file?

ChristopherChudzicki: Yes...I keep copying from terminal output and forgetting to change that back.

Do you want this as 1.1.1 in the changlong =/

jolyonb: No worries - I don't think this broke anything. We'll bundle some more work together and put out 1.1.1.

Issue #51, opened 2018-07-01
Check for collisions


jolyonb:
We should check to make sure there are no collisions between user-defined constants and variables (ie, no constant `k` and variable `k`, or worse, defining two constants/variables with the same name). We should also check that user-defined functions don't have duplicated names. We've decided that users should be able to overwrite default constants and functions, however.

ChristopherChudzicki:
We can check for variable--constant collisions and variable--variable collisions. I don't think we can check for constant--constant or function--function collisions without changing how constants and functions are declared because unlike variables, they are declared using dictionaries and Python dictionary literals permit duplicate keys and use the latest:

```python
{
    'a': 5,
    'a': 10
} ==
{
    'a': 10
}
# True
```

Regarding
> We've decided that users should be able to overwrite default constants and functions, however.

Another thing we could do is raise an error (ConfigCollisionError, or something) with a message like "ConfigCollisionError: Declaring variable 'i' will overwrite existing constant 'i'. To proceed anyway, add `suppress_warnings=True` to your grader configuration. Seems like it could be a useful pattern in other cases, too.

jolyonb:
Ah. Good point. Less work for us then! Let's just do variable-variable and variable-constant checks.

I'm happy to greenlight a warning message like that. Let's call it `ConfigCollisionWarning`. Putting in `suppress_warnings=True` will be annoying, but then again, that's kind of the point.

ChristopherChudzicki:
Ok! I'll do it now.

ChristopherChudzicki:
resolved by #53

PR #53, opened 2018-07-02
validate user-defined variables don't conflict, warn if override

ChristopherChudzicki:
This addresses #51:

- In FormulaGrader: if variables, numbered_vars, user_constants overlap, raise an error.
- In IntegralGrader: if variables and user_constants overlap, raise an error.

For both FormulaGrader and IntegralGrader, if any user-defined variable/function overrides a default, raise an error that can be suppressed with `suppress_warnings=True`.

### DIFF
--- a/docs/formula_grader.md
+++ b/docs/formula_grader.md
@@ -22,7 +22,7 @@ grader = FormulaGrader(
 
 Note that the `answers` parameter follows all of the usual allowances from ItemGrader.
 
-The variables need to have numbers randomly assigned to them. Each is sampled from a [sampling set](sampling.md), which is RealInterval() by default (random numbers from 1 to 3). A variety of different sampling sets are available, including random complex numbers. To specify the sampling set to use for a variable, use the `sample_from` key.
+The variables need to have numbers randomly assigned to them. Each is sampled from a [sampling set](sampling.md), which is RealInterval() by default (random numbers from 1 to 5). A variety of different sampling sets are available, including random complex numbers. To specify the sampling set to use for a variable, use the `sample_from` key.
 
 ````python
 grader = FormulaGrader(
@@ -98,55 +98,6 @@ grader = FormulaGrader(
 )
 ```
 
-### Blacklists and Whitelists
-
-You can disallow specific functions by adding them to the blacklist of functions as a list of disallowed function names. In the following example, `sin` is disallowed in correct answers.
-
-```python
-grader = FormulaGrader(
-    answers='sqrt(1 - cos(x)^2)',
-    variables=['x'],
-    sample_from={'x': [0, np.pi]},
-    blacklist=['sin']
-)
-```
-
-If you want to exclude everything except for a specific set of functions, instead use a whitelist. In the following example, the only allowed functions in correct answers are sin and cos.
-
-```python
-grader = FormulaGrader(
-    answers='sin(x)/cos(x)',
-    variables=['x'],
-    whitelist=['sin', 'cos']
-)
-```
-
-If you want to exclude all functions, use the following whitelist.
-
-```python
-grader = FormulaGrader(
-    answers='pi/2-x',
-    variables=['x'],
-    whitelist=[None]
-)
-```
-
-You cannot use a whitelist and a blacklist at the same time.
-
-
-### Required Functions
-
-You can specifically require certain functions to appear in the solution. Any solution that does not include all of these functions will generate an error message. To do this, specify a list of strings of function names that are required.
-
-```python
-grader = FormulaGrader(
-    answers='2*sin(theta)*cos(theta)',
-    variables=['theta'],
-    required_functions=['sin', 'cos']
-)
-```
-
-
 ### User Functions
 
 You can make user-defined functions available for students to use in their answers. To add user-defined functions, pass in a dictionary to the `user_functions` key as follows.
@@ -166,16 +117,6 @@ grader = FormulaGrader(
     answers="f''(x)",
     variables=['x'],
     user_functions={"f''": lambda x: x*x}
-)
-```
-
-If desired, you can override default functions.
-
-```python
-grader = FormulaGrader(
-    answers="x^2",
-    variables=['x'],
-    user_functions={"sin": lambda x: x*x}
 )
 ```
 
@@ -209,7 +150,7 @@ This allows you to grade mathematical expressions that involve unknown functions
 
 ## Constants
 
-By default, four constants are defined: `e`, `pi`, and `i`/`j`. You can define new constants by passing in a dictionary to `user_constants` as follows.
+By default, four constants are defined: `e`, `pi`, and `i=j=sqrt(-1)`. You can define new constants by passing in a dictionary to `user_constants` as follows.
 
 ```python
 grader = FormulaGrader(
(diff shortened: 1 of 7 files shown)


mitxgraders/formulagrader.py
```diff
@@ -219,6 +220,86 @@ def numbered_vars_regexp(numbered_vars):
...
+    ... }, keys)
+    True
+    """
+    dict_of_sets = {k: set(config[k]) for k in keys}
+    msg = "'{iter1}' and '{iter2}' contain duplicate entries: {duplicates}"
+
+    for k1, k2 in itertools.combinations(dict_of_sets, r=2):
+        k1, k2 = sorted([k1, k2])
+        duplicates = dict_of_sets[k1].intersection(dict_of_sets[k2])
+        if duplicates != set():
```
jolyonb: I think this line can just be `if duplicates:`

jolyonb:
This looks really good. You've done a good job here!

I'd like to update the documentation in a few more places.

* Somewhere in the root documentation for Graders, we should add a note about `suppress_warnings`
* For variables and constants, we should mention that you can override the defaults (do we even mention the defaults?)
* In numbered_vars, we mention that a specific variable takes precedence over a numbered variable. We should mention the suppress_warnings flag there.

I think we should also check to make sure that there are no collisions inside `variables` and `numbered_vars`. A quick and dirty check would be if `len(set(variables)) > len(variables)`. Pulling out the actual duplicates takes a bit more work.

docs/formula_grader.md
```diff
@@ -221,14 +162,35 @@ grader = FormulaGrader(
 )
 ```
 
-Constants are like variables that only ever have one value. You can overwrite constants if you like by passing in new definitions in user_constants, but we recommend against this. We also recommend ensuring that your constants and variables all have unique names. Variables will overwrite any constants that have the same name.
+Constants are like variables that only ever have one value.
 
 Note that by default, edX defines the constants `k`, `c` and `T`. In our experience, these constants lead to confusion when answers that should produce invalid variable errors instead simply eat a student's attempt. You can reintroduce them as necessary.
```
jolyonb: These are now gone!!!!! I think we can remove this paragraph.

ChristopherChudzicki: You mean edX removed `k`, `c`, and `T`? All of the constants, or just those? Were those the only ones?

jolyonb:
Line 58 of ` docs/formula_grader.md` needs updating to mention the warning suppression.

mitxgraders/formulagrader.py
```diff
@@ -397,8 +397,8 @@ def schema_config(self):
             Required('tolerance', default='0.01%'): Any(PercentageString, NonNegative(Number)),
             Required('metric_suffixes', default=False): bool,
             Required('samples', default=5): Positive(int),
-            Required('variables', default=[]): [str],
-            Required('numbered_vars', default=[]): [str],
+            Required('variables', default=[]): All([str], all_unique),
```
jolyonb: Now that's a good way to do it!!!

mitxgraders/helpers/validatorfuncs.py
```diff
@@ -137,6 +170,43 @@ def get_number_of_args(callable_obj):
 
     return num_args
 
+def dict_has_no_collisions(dict_of_iterables):
```
jolyonb: This looks nice; did you intend to use it somewhere?

ChristopherChudzicki: Oops. I'll delete it.

It's kind of a long story ... `dict_has_no_collisions` and `all_unique` were copy-pasted from a different branch and accidentally came together.

docs/formula_grader.md
jolyonb: Yup, they removed `k`, `c`, `Q` and `T` after I poked them :-) https://github.com/edx/edx-platform/pull/17368

ChristopherChudzicki:
> Somewhere in the root documentation for Graders, we should add a note about suppress_warnings
I added a brief note about validation and `suppress_warnings` to the main README.

> For variables and constants, we should mention that you can override the defaults (do we even mention the defaults?)
Yes, you mentioned the defaults in the README :)

I reorganized the FormulaGrader docs a little bit to group sampling functions and constants a bit closer together and to put blacklist/whitelist/required_functions/forbidden_strings all together. 

I think it's a bit clearer now that `suppress_warnings=True` is required for overriding either constants or functions.

Now, regarding:

> In numbered_vars, we mention that a specific variable takes precedence over a numbered variable. We should mention the suppress_warnings flag there.

and

> Line 58 of docs/formula_grader.md needs updating to mention the warning suppression.

Currently:

```python
grader = FormulaGrader(
    answers='a_{0} + a_{1}*x + 1/2*a_{2}*x^2',
    variables=['x', 'a_{0}'],
    numbered_vars=['a'],
    sample_from={
        'x': [-5, 5],
        'a': [-10, 10]
    }
)
```

does not raise an error, by it probably should, I suppose.

jolyonb:
Ah. I saw that you were checking to make sure that `numbered_vars` and `variables` didn't collide. I'm guessing that a variable `a` and a numbered variable `a` will trigger a collision warning, but a variable `a_{0}` and a numbered variable `a` won't. Honestly, I don't think it's worth worrying about collisions between variables and numbered variables, as the system has an unambiguous way to deal with that.

ChristopherChudzicki:
> Ah. I saw that you were checking to make sure that numbered_vars and variables didn't collide. I'm guessing that a variable a and a numbered variable a will trigger a collision warning, but a variable a_{0} and a numbered variable a won't.

Exactly—that much was easy to test for.

>  Honestly, I don't think it's worth worrying about collisions between variables and numbered variables, as the system has an unambiguous way to deal with that.

Yay :)

I'll just delete that doc line about q, c, etc and delete the unused function, then this is ready.

ChristopherChudzicki:
@jolyonb squashed and ready! 

Nice job on the constants with edX, they were annoying =) Are they still present for numericalresponse?

jolyonb:
No, they're gone completely :)

mitxgraders/formulagrader.py
```diff
@@ -410,8 +491,14 @@ def __init__(self, config=None, **kwargs):
         """
         super(FormulaGrader, self).__init__(config, **kwargs)
 
-        # finish validating blacklist/whitelist
+        # finish validating
         validate_blacklist_whitelist_config(self.config['blacklist'], self.config['whitelist'])
+        validate_no_collisions(self.config, keys=['variables', 'numbered_vars', 'user_constants'])
```
jolyonb: You need to take `numbered_vars` out of this collision validation, per our discussion.

jolyonb:
One last little thing sorry...

mitxgraders/formulagrader.py
ChristopherChudzicki: Wait, I thought we wanted to keep that. That ensures
```python
grader0 = FormulaGrader(
    variables=['a'],
    numbered_vars=['a']
)
```
is invalid. That should be invalid, and it's easy to validate. That will give you a non-suppressible error.

In contrast:
```python
grader1 = FormulaGrader(
    variables=['a_{1}'],
    numbered_vars=['a']
)
```
is valid and raises no error (suppressible or otherwise). In an ideal world, grader1 would raise a suppressible error, but we decided that is not worth it.

jolyonb: I don't think there's anything wrong with `a` being a variable and a numbered variable, because that allows variables `a` and `a_{n}` in the same problem. It's a bit unusual, but there's no issues internally for us?

jolyonb: one final commit?

ChristopherChudzicki: Ugh. I’m not very on top of it today, apparently. 

  I’m away from my computer now, but I will fix it tonight or tomorrow morning.

ChristopherChudzicki:
@jolyonb I guess I just forgot to push the last commit, which is now pushed. This should in fact be ready.

PR #54, opened 2018-07-03
Removing __pycache__ files from zip

jolyonb:
For some reason, I have python3 bytecode in __pycache__ folders. This updates the zip routine to remove these folders.

Perhaps there's a better way to make the zip file that only includes `.py`, `.pyc` and `LICENSE` files recursively, so that the zip doesn't pick up anything that's in `.gitignore`?


ChristopherChudzicki:
Do you want to remake the zip after #53 ?

jolyonb:
Done. Have to remember to rebase to get changes from master, not merge...

PR #55, opened 2018-07-03
Fixes voluptuous error module loading

jolyonb:
It needed a local reference (I believe it's a python3 thing that updated referencing requirements).

Also updated the zip file again - MITx support suggested that the pyc files we generate might be mismatching with the version that edX runs on, which could cause errors.


ChristopherChudzicki:
Re
> [Voluptuous] needed a local reference (I believe it's a python3 thing that updated referencing requirements).

Ah. Incidentally, those imports were originally `from voluptuous.error import ... ` when I copied voluptuous from https://github.com/alecthomas/voluptuous into mitxgraders, but that didn't work so I believe I changed it to relative imports.

I don't know that the best way to include external libraries like voluptuous into our own. This doesn't seem ideal, but also not bad since voluptuous only has 1 import from within itself, all others are standard library.

jolyonb:
Yeah, I don't know what the best way to have somebody else's package contained in your own package is. There's probably a best practice that actually works. I suspect it involves moving voluptuous out of our package entirely? Probably worth looking into...

ChristopherChudzicki:
@jolyonb Well, I don't know how common this situation actually is in the Python community. I feel like the normal thing would be to just put it in `requirements.txt`, but that doesn't work for customresponse on edX.

PR #56, opened 2018-07-03
Moving voluptuous outside of mitxgraders

jolyonb:
What it says on the tin. Turned out to be really straightforward. I've included a small amount of linting here too.


ChristopherChudzicki:
Sure, we can do it that way! I guess keeping 1 folder inside python_lib isn't really necessary. Doesn't change installation process if ours is your only python_lib.zip. Readme says "merge ours with yours", so that's still correct.

Do you think it's worth having a `try/except` in mitxgraders/__init__.py for voluptuous, with a custom error message? Like "ImportError: External dependency 'voluptuous' not found. See [link to FAQ]."? But I'm not sure that's necessary—happy to merge as is.

jolyonb:
That's a nice touch. Added.

mitxgraders/__init__.py
```diff
@@ -12,6 +12,13 @@
 # Add the current version
 from mitxgraders.version import __version__
 
+# Add voluptuous
+try:
+    import voluptuous
+except ImportError:
+    raise ImportError("External dependency 'voluptuous' not found;"
```
ChristopherChudzicki: Oh no, the coverage! 

So after a bit of research, we can test those lines with:
```python
import sys
import mock
from pytest import raises
import mitxgraders

def test_voluptuous_import_error_message():
    with mock.patch.dict(sys.modules, {'voluptuous': None}):
        msg = ("External dependency 'voluptuous' not found;"
               " see https://github.com/mitodl/mitx-grading-library#faq")
        with raises(ImportError, match=msg):
            # importing is not good enough to raise the error; need to reload
            # because conftest.py already imported voluptuous
            reload(mitxgraders)
```
(I installed `mock` a while back for testing comparer functions anyway).

This was perhaps more effort than it was worth, but fun to learn :).

jolyonb: I needed to import reload from imp, but other than that, seems to work!

Issue #43, opened 2018-06-30
Exponentiation issue in calc.py


jolyonb:
There is an issue with calc.py whereby exponentiation doesn't parse in the correct order. I can't remember exactly what the issue is, but I remember that we haven't fixed it.

jolyonb:
Ok, good to have this documented. I looked at calc.py to see why this is happening... and it's baked somewhat strongly into the tree parser. We'd have to refactor that algorithm significantly :-(

ChristopherChudzicki:
Resolved by #59

Issue #42, opened 2018-06-30
Multivariable functions


jolyonb:
We should be able to allow user-defined functions that take in multiple variables. Eg, `f(x, y)`. This will require some extra parsing in `calc.py`, and will also require some extra checking to ensure that the correct number of arguments are being passed to any given function.

ChristopherChudzicki:
Resolved by #59

PR #59, opened 2018-07-05
Implementing multi-argument functions and fixing exponentiation issue

jolyonb:
Addresses issues #42 and #43 by rewriting calc.py to fix exponentiation problem (introduced "negation" operation) and allow multi-argument functions (introduced "arguments" operation). Also did general tidying of pyparsing script and some refactoring of calc.py, along with a little linting. Large changes to your `get_number_of_args` function were necessary to pull this off.


jolyonb:
I'm quite proud of this one. Let me know what you think!

mitxgraders/helpers/calc.py
```diff
@@ -549,23 +448,331 @@ def check_variables(self, valid_variables, valid_functions):
...
+            parse_result: A list containing zero or more "-" strings, followed by a number.
+            ["-", "-", "-", a] = -a
+            ["-", "-", a] = a
+
+        Usage
+        =====
+        >>> parser = FormulaParser("1", {"%": 0.01})
+        >>> parser.eval_negation([2])
+        2
+        >>> parser.eval_power(["-",2])
```
ChristopherChudzicki: Should this (and the next few) be `eval_negation`? (Or, `eval_sign` if you take the suggestion above.)

mitxgraders/helpers/calc.py
```diff
@@ -447,99 +324,121 @@ def parse_algebra(self):
...
-        prod_term = Group(prod_term)("product")
+        # Define multiplication and division
+        product = Group(parallel +
+                        ZeroOrMore((Literal('*') | Literal('/')) + parallel))("product")
 
-        sum_term = Optional(plus_minus) + prod_term + ZeroOrMore(plus_minus + prod_term)  # -5 + 4 - 3
-        sum_term = Group(sum_term)("sum")
+        # Define sums and differences
+        # Note that leading - signs are treated by negation
+        sumdiff = Group(Optional(plus) + product +
```
ChristopherChudzicki: | Example | Expression   	| master parses?  	| multivar parses?  |
|---      |---	          |---	              |---	      |
| A        | 1+2   	      | yes   	          | yes       |
| B       | 1+-2  	      | yes  	            | yes  	    |
| C       | 1-+2  	      | yes  	            | no  	    |
| D       | 1++2  	      | yes^1             | no  	    |
| E       | 1+++2  	      | no  	            | no  	    |
| F       | ---2  	      | no  	            | yes  	    |
| G       | --+2  	      | no  	            | no  	    |

1: '++2' behaves similarly, both are parsed roughly as:
```
<sum>
  <number>1</number>
  +
  <number>
   +1 
  </number>
</sum>
```

Examples C through G are certainly a bit contrived, but I don't think they're too ridiculus.

I'm actually pretty confused about why the `ZeroOrMore` in line \#382 doesnt' permit `1+++2` (on either this branch or master).

**Suggestion:** Change `negation` to `sign` or `unary_sign`, and change `eval_negation` to `eal_sign`. (I can commit my local version of this if you want, or you can write it.)

tests/test_base.py
```diff
@@ -10,11 +10,13 @@
 from pytest import raises, approx
 from voluptuous import Error, Invalid, truth
 import mitxgraders
-from mitxgraders import ListGrader, StringGrader, ConfigError, FormulaGrader, CalcError, __version__
+from mitxgraders import (ListGrader, StringGrader, ConfigError, FormulaGrader,
+                         CalcError, RandomFunction, __version__)
 from mitxgraders.helpers import validatorfuncs
-from mitxgraders.helpers.calc import evaluator, UnableToParse, UndefinedVariable
```
ChristopherChudzicki: A bunch of stuff in `test_base.py` testing stuff from mitxgraders.helpers. Maybe we should just make `test/helpers/test_calc.py` and `test/helpers/test_validatorfuncs.py`?

mitxgraders/helpers/calc.py
```diff
@@ -29,48 +26,44 @@
     alphas,
     nums,
     stringEnd,
-    ParseException
+    ParseException,
+    delimitedList
```
ChristopherChudzicki: sweet :)

tests/test_base.py
jolyonb: I've refactored it out as suggested.

mitxgraders/helpers/calc.py
jolyonb: I know, right? I was quite happy not to have to construct it in terms of more low-level functionality. And it will make doing vectors/matrices quite straightforward, too!

mitxgraders/helpers/calc.py
jolyonb: So, I deliberately treated + and - differently here, basing things on how google parses math input. I don't want to accept + as part of a number, like `1-+2` would require. It's ok for - to be part of a number, which is why `1-------2` works - you just keep on adding brackets around the 2 with an ever-increasing number of minus signs. 

I always thought it was strange that the old parser allowed for two pluses in a row, but not three, also. The reason is that it allowed for a single plus to be a part of a number, and then a second plus to be a part of a sum. 

I think we should discourage inputs like your D. I also don't like C, E and G for the same reason. I'm happy with B, and I'm willing to tolerate F.

Note that like 382 contains `ZeroOrMore(plus_minus + product)`, so you must have an expression after a +/- sign.

jolyonb:
Ok, made some minor modifications, and responded to your more major point.

I'm pretty happy that that's all you were able to find, given my 4:30 am coding!

mitxgraders/helpers/calc.py
ChristopherChudzicki: > I don't want to accept + as part of a number
Still thinking about what you said, but to be clear my suggestion was

```python
# replace negation with this:
sign = Group(ZeroOrMore(plus_minus) + power)("sign")

# and later, replace eval_number with
def eval_sign(self, parse_result): 
        num_negations = len([char for char in parse_result[:-1] if char == '-' ])
        num = parse_result[-1]
        return num * (-1)**(num_negations)
```

jolyonb: Yup, I understood the suggestion.

Do you agree with me that we shouldn't allow C/D/E/G and we need to allow B?

I'd be willing to get rid of F.

To add another case, I think `1--2` needs to be allowed, but `1---2` can probably go.

I think these can be accomplished by changing `ZeroOrMore` to `Optional` in the negation statement.

ChristopherChudzicki: Yes, I agree those are bad. Thinking about it more—I like rejecting them as unparseable: I think it would be easy to miscount the number of signs.

I agree Optional is a good choice.

But really, I guess it doesn't matter. I just checked some 2015 data, and in 400,000 submissions I found about 25 `+-`, 10 `-+`, and 7 `++`. (Not all of the 400,000 were FormulaResponse submissions, though. Still seems rare.)

Issue #60, opened 2018-07-05
Reduce unncessary nodes in parse tree.


ChristopherChudzicki:
The parser in `calc.py` attaches `sum/product/parallel/power` to *every* set of parentheses, even if there's nothing to add/multiply/etc. For example, `((2))` parses as:

```xml
<sum>
  <product>
    <parallel>
      <negation>
        <power>
          <atom>
            <parentheses>
              <sum>
                <product>
                  <parallel>
                    <negation>
                      <power>
                        <atom>
                          <parentheses>
                            <sum>
                              <product>
                                <parallel>
                                  <negation>
                                    <power>
                                      <atom>
                                        <number>
                                          <ITEM>2</ITEM>
                                        </number>
                                      </atom>
                                    </power>
                                  </negation>
                                </parallel>
                              </product>
                            </sum>
                          </parentheses>
                        </atom>
                      </power>
                    </negation>
                  </parallel>
                </product>
              </sum>
            </parentheses>
          </atom>
        </power>
      </negation>
    </parallel>
  </product>
</sum>
```

it would be nicer (more readable during debugging) we could through away grouped nodes with only 1 child.

jolyonb:
I would love this. I couldn't figure out how to get pyparsing to do it during my refactoring though, so I went for the method I knew would work. I think this would also significantly speed up both the parsing and evaluation time.

jolyonb:
Addressed in #62.

PR #62, opened 2018-07-05
simplify the generated parse tree

ChristopherChudzicki:
This addresses #60 by removing unnecessary groups (`atom` and `parentheses`, whose eval functions just returned the group children) and by conditionally grouping sum/product/parallel/power/negate only if there are children.

This generates much cleaner parse trees. For example, `3^2+(2)` generates

```xml
<sum>
  <power>
    <number>
      <ITEM>3</ITEM>
    </number>
    <number>
      <ITEM>2</ITEM>
    </number>
  </power>
  <ITEM>+</ITEM>
  <number>
    <ITEM>2</ITEM>
  </number>
</sum>
```

whereas the old version would generate:

```xml          
<sum>
  <product>
    <parallel>
      <negation>
        <power>
          <atom>
            <number>
              <ITEM>3</ITEM>
            </number>
          </atom>
          <atom>
            <number>
              <ITEM>2</ITEM>
            </number>
          </atom>
        </power>
      </negation>
    </parallel>
  </product>
  <ITEM>+</ITEM>
  <product>
    <parallel>
      <negation>
        <power>
          <atom>
            <parentheses>
              <sum>
                <product>
                  <parallel>
                    <negation>
                      <power>
                        <atom>
                          <number>
                            <ITEM>2</ITEM>
                          </number>
                        </atom>
                      </power>
                    </negation>
                  </parallel>
                </product>
              </sum>
            </parentheses>
          </atom>
        </power>
      </negation>
    </parallel>
  </product>
</sum>
```

### DIFF
--- a/mitxgraders/helpers/calc.py
+++ b/mitxgraders/helpers/calc.py
@@ -222,7 +222,6 @@ def evaluator(formula, variables, functions, suffixes, allow_vectors=False):
     # Return the result of the evaluation, as well as the set of functions used
     return result, math_interpreter.functions_used
 
-
 class FormulaParser(object):
     """
     Parses a mathematical expression into a tree that can subsequently be evaluated
@@ -245,14 +244,13 @@ class FormulaParser(object):
             'variable': self.eval_var,
             'arguments': self.eval_arguments,
             'function': self.eval_function,
-            'parentheses': self.eval_parens,
             'vector': self.eval_vector,
-            'atom': self.eval_atom,
             'power': self.eval_power,
             'negation': self.eval_negation,
             'parallel': self.eval_parallel,
             'product': self.eval_product,
-            'sum': self.eval_sum
+            'sum': self.eval_sum,
+            'parentheses': lambda tokens: tokens[0] # just get the unique child
         }
         self.vars = {}
         self.functions = {}
@@ -274,6 +272,20 @@ class FormulaParser(object):
         When vector input is used, mark it in `vectors_used`.
         """
         self.vectors_used = True
+
+    @staticmethod
+    def group_if_multiple(name):
+        """
+        Generates a parse action that groups ParseResults with given name if
+        ParseResults has multiple children.
+        """
+        def _parse_action(tokens):
+            """Wrap children in a group if there are multiple"""
+            if len(tokens) > 1:
+                return ParseResults(toklist=[tokens], name=name)
+            return tokens
+
+        return _parse_action
 
     def parse_algebra(self):
         """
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki:
@jolyonb What do you think about this?

There is an argument to be made that the original version better represents what our parser is actually doing, and so might be more useful for debugging. But the original version is also unwieldy.

Note: I also doubt this will improve performance significantly if at all. It's the parsing that is slow, and I really haven't changed the parsing much, just what is stored.

The only sense in which this might be faster is that before, each evaluation call had to deal with a bunch of sum/diff/power that had exactly 1 child. Now those only-child cases are dealt with at parse time instead of at evaluation time. But they were probably fast to deal with in the first place.

jolyonb:
This is actually a really clever way to do it. I completely approve. Some minor nits incoming.

mitxgraders/helpers/calc.py
```diff
@@ -275,6 +272,20 @@ def vector_parse_action(self, tokens):
         """
         self.vectors_used = True
 
+    @staticmethod
+    def group_if_multiple(name):
+        """
+        Generates a parse action that groups ParseResults with given name if
+        ParseREsults has multiple children.
```
jolyonb: `ParseResults`, not `ParseREsults`. I told you there were minor nits!

mitxgraders/helpers/calc.py
```diff
@@ -363,9 +374,7 @@ def parse_algebra(self):
         function.setParseAction(self.function_parse_action)
 
         # Define parentheses
-        parentheses = Group(Suppress("(") +
-                            expr +
-                            Suppress(")"))("parentheses")
+        parentheses = Suppress("(") + expr + Suppress(")")
```
jolyonb: Can we actually keep parentheses around, please? I like having these with an explicit grouping tag in debug mode.

mitxgraders/helpers/calc.py
```diff
@@ -610,22 +621,6 @@ def eval_function(self, parse_result):
                        "Its input does not seem to be in its domain.").format(name=name)
                 raise FunctionEvalError(msg)
 
-    def eval_parens(self, parse_result):
```
jolyonb: Honestly, this can be made into a lambda function in the dictionary above.

jolyonb:
`eval_arguments` and `eval_var` should also probably just be lambda functions, given that they're one-liners.

ChristopherChudzicki:
@jolyonb I restored the parentheses. I don't mind having simple things as named functions, then they show up in tracebacks nicer. But for some reason I did make eval_parentheses a lambda. I am inconsistent. 

Anyway, you can merge or lambdify those other ones if you want :)

jolyonb:
Haha, ok ;-) Very nice. Merging now!

PR #63, opened 2018-07-05
Tidying up calc.py

jolyonb:
This does a bunch of tidying in calc.py.

* `evaluator` now has default arguments for variables, functions and suffixes.
* Made all negative signs limited to just one, and made a bunch of tests to ensure compliance.
* Moved one-line pyparsing actions to lambda functions.
* Cleaned up `dump_parse_result`
* Gave names to all symbols in the parser tree, so that you don't see any more `ITEM` tags in the debug dump. Eg., `1+2^-2*sin(x)` now parses to the following debug tree:

```XML
<sum>
  <number>
    <num>1</num>
  </number>
  <op>+</op>
  <product>
    <power>
      <number>
        <num>2</num>
      </number>
      <op>-</op>
      <number>
        <num>2</num>
      </number>
    </power>
    <op>*</op>
    <function>
      <funcname>sin</funcname>
      <arguments>
        <variable>
          <varname>x</varname>
        </variable>
      </arguments>
    </function>
  </product>
</sum>
```

I was trying to get the variable tags to work as `<variable>x</variable>`, but I couldn't seem to get that to happen...

### DIFF
--- a/mitxgraders/helpers/calc.py
+++ b/mitxgraders/helpers/calc.py
@@ -30,6 +30,9 @@ from pyparsing import (
     delimitedList
 )
 from mitxgraders.helpers.validatorfuncs import get_number_of_args
+from mitxgraders.helpers.mathfunc import (DEFAULT_FUNCTIONS,
+                                          DEFAULT_SUFFIXES,
+                                          DEFAULT_VARIABLES)
 
 class CalcError(Exception):
     """Base class for errors originating in calc.py"""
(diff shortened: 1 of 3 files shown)


jolyonb:
@ChristopherChudzicki This is ready for proofing now.

mitxgraders/helpers/calc.py
```diff
@@ -164,7 +167,11 @@ def get_parser(self, formula, suffixes):
 # The global parser cache
 parsercache = ParserCache()
 
-def evaluator(formula, variables, functions, suffixes, allow_vectors=False):
+def evaluator(formula,
+              variables=DEFAULT_VARIABLES,
```
ChristopherChudzicki: Just confirming: You don't mind the mutable default argument because nothing mutates this below? 

Otherwise, 👍 So if you're fine with it, feel free to merge.

jolyonb: I'm fine with it. An alternative is to have them set to `NONE`, check that and then default to the variables, but given that nothing mutates it anyway, I don't think anything is going to go wrong.

jolyonb: And this is just a nicer implementation, too...

Issue #46, opened 2018-06-30
Correlated Grading


jolyonb:
When we have a ListGrader that contains multiple inputs (lowest-level of nesting), the grading for an input should be able to depend on the content of another input. A simple example would be "enter any number here", then "enter its square here". Because the first entry can accept basically anything, that would require a custom comparer. The second entry would also require a custom comparer. The easiest implementation of this would be to allow a custom comparer to have a special parameter that evaluates another input. We'd just need to be careful so as to have no circular references!

jolyonb:
Resolved by #80.

PR #65, opened 2018-07-06
Implementing correlated grading

jolyonb:
This PR adds a `dependent_input` config variable to each `ItemGrader` class that specifies which inputs are required to perform the check for that `ItemGrader`. When a lowest-level `ListGrader` has ordered input with individual subgraders, it reads the `dependent_input` list from each subgrader. Those with empty lists, it evaluates. Those that have dependencies, it passes through to the `check` function as a keyword argument.

I've implemented `dependent_input` for `FormulaGrader` only (and `NumericalGrader` by subclassing). For `FormulaGrader`, the dependencies are computed after variables have been sampled. The dependencies are labelled `input_n`, and added to the variables list before computing any `comparer_params`. This means that `input_n` dependencies can be used in the answer strings for such graders without needing a custom comparer. The `input_n` variables are scrubbed from the variables list before the student input is checked, so students cannot take advantage of it.

This was actually surprisingly straightforward to implement. So far however, this is just the raw implementation. This still needs the following:

- [x] Implementation
- [ ] Validation
- [ ] Documentation
- [ ] Course example
- [ ] Tests

Addresses issue #46.

### DIFF
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -16,6 +16,7 @@ import abc
 from voluptuous import Schema, Required, All, Any, Range, MultipleInvalid
 from voluptuous.humanize import validate_with_humanized_errors as voluptuous_validate
 from mitxgraders.version import __version__
+from mitxgraders.helpers.validatorfuncs import all_unique
 
 class ConfigError(Exception):
     """Raised whenever a configuration error occurs"""
@@ -92,13 +93,15 @@ class AbstractGrader(ObjectWithSchema):
         })
 
     @abc.abstractmethod
-    def check(self, answers, student_input):
+    def check(self, answers, student_input, **kwargs):
         """
         Check student_input for correctness and provide feedback.
 
         Arguments:
             answers: The expected result(s) and grading information
             student_input: The student's input passed by edX
+            **kwargs: Anything else that has been passed in to the grader
+                      (used for passing dependent_inputs)
         """
 
     def __call__(self, expect, student_input):
@@ -227,7 +230,9 @@ class ItemGrader(AbstractGrader):
         schema = super(ItemGrader, self).schema_config
         return schema.extend({
             Required('answers', default=tuple()): self.schema_answers,
-            Required('wrong_msg', default=""): str
+            Required('wrong_msg', default=""): str,
+            # dependent_input is used by ListGrader for correlated grading
+            Required('dependent_input', default=[]): All([int], all_unique)
         })
 
     def schema_answers(self, answer_tuple):
@@ -298,7 +303,7 @@ class ItemGrader(AbstractGrader):
         """
         return Schema(str)(expect)
 
-    def check(self, answers, student_input):
+    def check(self, answers, student_input, **kwargs):
         """
         Compares student input to each answer in answers, using check_response.
         Computes the best outcome for the student.
@@ -306,6 +311,8 @@ class ItemGrader(AbstractGrader):
         Arguments:
             answer: A tuple of answers to compare to, or None to use internal config
             student_input (str): The student's input passed by edX
+            **kwargs: Anything else that has been passed in to the grader
+                      (used for passing dependent_inputs)
         """
         # If no answers provided, use the internal configuration
         answers = self.config['answers'] if answers is None else answers
@@ -324,7 +331,7 @@ class ItemGrader(AbstractGrader):
             raise ConfigError(msg.format(type(student_input)))
 
         # Compute the results for each answer
-        results = [self.check_response(answer, student_input) for answer in answers]
+        results = [self.check_response(answer, student_input, **kwargs) for answer in answers]
 
         # Now find the best result for the student
         best_score = max([r['grade_decimal'] for r in results])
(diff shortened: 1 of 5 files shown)


ChristopherChudzicki:
Happy to pick this up after at least an initial PR for the matrix stuff.

ChristopherChudzicki:
@jolyon This is really cool works great, even with nested ListGraders. 

I've been thinking about correlated grading and correlated inputs for ListGrader, and it strikes me that they are actually two separate issues:

- **correlated grading**: For ItemGraders inside a ListGrader, provide each ItemGrader with the other submitted answer.
- **correlated input** For ItemGraders inside a ListGrader, when a student enters their input for one box, allow them to reference their inputs to other boxes. 

Correlated input sounds interesting but hard. See further comments below.

### Correlated Grading

I think we can get a lot of mileage out of correlated grading and implement correlated grading more simply, without checking for circular dependencies or requiring subgraders. For example:

```python
grader = ListGrader(
    answers=['a', 'input_1^2', '(input_1 + input_2)^2'],
    subgraders=FormulaGrader(variables=['a'])
)

grader(0, ['a', '2*a', '3*a']) # This input is not correct!
```
In this example, when the ListGrader calls FormulaGrader's check method, it would pass an additional keyword argument every time:
```python
sibling_values = ['a', '2*a', '3*a']
# FormulaGrader would be responsible for turning this into 
# {'input_1': 'a', 'input_2': '2*a', 'input_3': '3*a'}
```

Note in particular that circular references are not a problem. For example:
```python
grader = ListGrader(
    answers=['input_2', 'input_1'],
    subgraders=FormulaGrader(variables=['a'])
)
grader(0, ['a', 'a'])['ok'] # True !
grader(0, ['10*a', '7*a + 3*a'])['ok'] # True !
grader(0, ['', ''])['ok'] # Empty strings are evaluated as 'nan', so I don't know what would happen here...
```
Well, ok. Circular references would allow authors to do stupid things like `answers=['input_2', 'input_1 + 1']`, but I don't think we should worry about that.

One nice thing about this approach is that it would let you use a single FormulaGrader() for your whole list, so you wouldn't need to re-specify variables, etc.

### Correlated Input
Allowing students to reference their own inputs sounds harder and would definitely require sorting dependencies. I think it also only makes sense for FormulaGrader. I would suggest implementing it through DependentSampler. Soemthing like:

```python
# Problem: List the 2nd through 4th terms in a geometric sequence with common ratio r and initial value a0 = 2.
grader = ListGrader(
  # answers are named 'a2', 'a3', 'a4'
  answers=['2r','2r^2', '2r^3'],
  subgraders=FormulaGrader(
    variables=['a2','a3','a4', 'r'],
    sample_from={
      'a2': DependentSample(depends=['input_1'], formula='input_1'),
      'a3': DependentSample(depends=['input_2'], formula='input_2')
    }
  )
)
```
This would require some changes to how DependentSampler evaluates.

Overall, I think correlated inputs are not worth it without a compelling reason.

(BTW: If we ever do need to change how DependentSampler works, or sort dependencies in some other area, I think we should use a package like [toposort](https://pypi.org/project/toposort/) to do this for us.)

jolyonb:
Indeed, passing all `siblings` and exposing these as variables sounds like a clean way to do it. You can get away without doing the circular reference check, but note in my example that it only exists to ensure that the author hasn't set one up - I could have just evaluated them all anyway. Up to you.

I'm not interested in correlated input at this stage. Students can do their own damn calculations ;-)

ChristopherChudzicki:
OK. I'm going out for a bit, but I'll refactor it that way later.

ChristopherChudzicki:
@jolyonb Travis seems not to care if coverage decreases anymore. Did you change that on purpose?

ChristopherChudzicki:
@jolyonb Cool, I'll fix those typos!

One of the things on your checklist for #65 was validation. Two comments about this:

1. I suspect the most common configuration error would be something like:
```python
grader = listGrader(
  answers=['1', 'sibling_1 + 1'],
  subgraders=FormulaGrader()
) # instantiated OK
grader(None, ['1', '2']) # throws InvalidInput: sibling_1 undefined
```
The issue is that the author forgot to declare the ListGrader as being ordered. FormulaGrader validates itself without any knowledge of sibling variables or the ListGrader's answers, so the error doesn't arise until check time.
  - This is sort of present in your original implementation (same error if author tries to use `input_k` variables without declaring `dependent_inputs`). You avoided number 2 by checking for cyclic dependencies in `dependent_inputs`.

2. Authors can set up ridiculous things like
```python
grader = ListGrader(
    answers=['sibling_1 + 1'],
    subgraders=FormulaGrader(),
    ordered=True
)
```
Nothing would ever be marked correct, since the answer cannot be 1 bigger than itself.
  - I am not very worried about this issue.

Mostly: **I think avoiding the extra key (`dependent_inputs`) and allowing reuse of the subgrader is worth the trouble of these less-than-helpful error messages.**

I'm also not super worried about it for what I suspect will be an uncommonly used feature.

A hacky but fast resolution to the first issue would be to raise a special error if the undefined variable contains `'siblings'`. Students *might* see this, but why would they be entering 'sibling_k' as a variable name....

Anyway, I think the PR is fine, but I wanted you to be aware of those issues.

jolyonb:
That's a coveralls thing. I don't know why it isn't complaining...

jolyonb:
What I was thinking about for validation was requiring ordered and a list of subgraders in order to be using the `dependent_input` key.

I think at the end of the day, issue 1 is ok -- it's not like during config we validate that the expression given by the instructor can even evaluate.

Issue 2 -- I agree that we shouldn't worry about it. It's equivalent to the instructor putting a wrong answer in anyway.

jolyonb:
Superseded by #80.

PR #70, opened 2018-07-09
Matrix sampling

ChristopherChudzicki:
This PR provides vectors/matrices/tensors for `calc.py` and introduces a few structural changes. In particular:

- A new base exception class `StudentFacingError`. When a `StudentFacingError` is raised within calc.py, it is presented to the student as-is. That is, these errors are intended to be read by students. My chief goal in introducing StudentFacingError is to move error-handling to the objects that generate errors instead having that responsibility fall on calc.py.
  - @jolyonb There is possibly some overlap between `StudentFacingError` and `InvalidInput`? 
- A new subclass of `numpy.ndarray` called `MathArray`. MathArrays are like arrays, but behave how a mathematician would expect.
   - For details, see MathArray's docstring.
- A new class `IdentityMultiple` that represents a multiple of the identity operator. Essentially, an identity matrix with unspecified dimension.

For now, vectors/matrices are enabled through a configuration option `max_array_dim` in FormulaGrader. See [test_formulagrader.py](https://github.com/mitodl/mitx-grading-library/compare/master...ChristopherChudzicki:matrix-sampling?expand=1#diff-4c1c44adafdb9ad1ab358dab3a3166b9R588) for a neat example.

@jolyonb This does **not** include documentation (besides docstrings) or course examples. I think some public API issues (e.g., should we have a separate ArrayFormulaGrader and configuration key names) are still undecided, but I thin the underlying approach is worth reviewing now.


jolyonb:
Ok, I'll take a look at this now, and we can do a later PR for documentation/course examples. I expect those to be somewhat hefty, so it's reasonable to split them off.

Can you fix the failing tests before I start review?

mitxgraders/helpers/math_array.py
```diff
@@ -0,0 +1,371 @@
...
+        """
+        if self.ndim == 0:
+            if isinstance(other, Number):
+                return robust_pow(self.item(), other)
+            elif isinstance(other, MathArray):
+                return other.__rpow__(self.item())
+            else:
+                raise TypeError("Cannot raise a scalar to a {type} power".format(type=type(other)))
+
+        elif not self.ndim == 2:
```
jolyonb: Can we add in a special case for `self.ndim == 1` and `other == 2` to take the norm of a vector?

ChristopherChudzicki:
@jolyonb I'll look into the failing test. It's some stupid error with the debug log. I'm always worried about rounding issues there. Everything else is passing, so I think it's safe to look.

jolyonb:
I think `InvalidInput` has a huge amount of overlap with `StudentFacingError`. I think having a base class of exceptions that go straight through is a great idea though, and we should refactor `InvalidInput` appropriately. One possible catch: if such an error is raised on an instructor's input, do we need to recast it as a `ConfigError`?

jolyonb:
Ok, I'll start working my way through this. Might take me a couple of days - this is definitely the biggest PR we've had since v1.0, and I can only think of a couple of challengers even before that!

mitxgraders/helpers/math_array.py
ChristopherChudzicki: I think we should use `abs(v)` or `norm(v)` for that. (Which reminds me, I'm pretty sure I made `abs` a scalar-only function. I should fix that.)

(The only course I ever took where `v^2` meant `v dot v` for vectors was graduate stat mech ...)

jolyonb:
I just went through the commit log and clicked on some of the commits I knew were large - I only caught one that went over a thousand (major restructuring of ListGrader), and none that challenged this remotely. The only possible larger ones were before I got involved!

mitxgraders/baseclasses.py
```diff
@@ -26,6 +26,10 @@ class InvalidInput(Exception):
     """Raised whenever user input is invalid"""
     pass
 
+class StudentFacingError(Exception):
```
ChristopherChudzicki: > I think InvalidInput has a huge amount of overlap with StudentFacingError. I think having a base class of exceptions that go straight through is a great idea though, and we should refactor InvalidInput appropriately. One possible catch: if such an error is raised on an instructor's input, do we need to recast it as a ConfigError?

That was my inclination too. I think we should do that refactoring, but it might be best to do in a subsequent PR—seems like I might end up changing a bunch of files that aren't really related to vectors/matrices. But let me know if you feel otherwise.

mitxgraders/helpers/math_array.py
jolyonb: Really? `v^2 = v dot v` is really very common! If `abs` plays the same role though, that's fine by me.

mitxgraders/baseclasses.py
jolyonb: I agree - let's do that separately.

ChristopherChudzicki:
Re
> Ok, I'll start working my way through this. Might take me a couple of days - this is definitely the biggest PR we've had since v1.0, and I can only think of a couple of challengers even before that!
and
> I just went through the commit log and clicked on some of the commits I knew were large - I only caught one that went over a thousand (major restructuring of ListGrader), and none that challenged this remotely. The only possible larger ones were before I got involved!

Yeah, that's why I wanted to get the review rolling :) Take your time! As for the lines. A lot of that is probably doc and tests.

jolyonb:
Hmm. For some reason, Travis is only giving one check here, rather than the two it normally does. Also, coveralls isn't working. I suspect that coveralls was related to the other Travis run... I'll look into it.

ChristopherChudzicki:
@jolyonb Tests fixed!

ChristopherChudzicki:
@jolyonb I think coveralls did run?

jolyonb:
Huh, so it did. It just took it's time... Ok, well, only one Travis check now :-)

mitxgraders/helpers/calc.py
```diff
@@ -529,7 +531,7 @@ def eval_number(self, parse_result):
         """
         result = float(parse_result[0])
         if len(parse_result) == 2:
-            result *= self.suffixes[parse_result[1]]
```
jolyonb: Was there some weirdness with the magic functions to make this work that precluded its use? It's going to make future debugging really confusing if so!

mitxgraders/helpers/math_array.py
```diff
@@ -0,0 +1,371 @@
...
+
+    >>> map(is_number_zero, [0, 0.0, 0j])
+    [True, True, True]
+    >>> is_number_zero(np.matrix([0, 0, 0]))
+    False
+    """
+    return isinstance(value, Number) and value == 0
+
+def is_scalar_array_zero(value):
+    return isinstance(value, MathArray) and value.ndim == 0 and value.item() == 0
```
jolyonb: Docstring?

mitxgraders/helpers/math_array.py
```diff
@@ -0,0 +1,371 @@
...
+
+    Differences from numpy.ndarray:
+    ==============================
+        - Multiplication:
+            - scalar * array = scales elements of array;
+            - array * scalar = scales elements of array;
+            If shapes are correct:
+            - vector * vector = number
+            - matrix * vector = vector
+            - vector * vector = vector
```
jolyonb: This line looks wrong? Was one of those meant to be matrix?

mitxgraders/helpers/math_array.py
```diff
@@ -0,0 +1,371 @@
...
+            - matrix * matrix = matrix
+            Any multiplication involving tensors and arrays of dimension >1 is not supported.
+        - powers are only allowed for square matrices and integer-like exponents.
```
jolyonb: non-negative integer-like?

mitxgraders/helpers/math_array.py
```diff
@@ -0,0 +1,371 @@
...
+
+    def __new__(cls, input_array):
+        """
+        Creates a new MathArray Instance.
+
+        Note: The ndarray constructor is very low-level and apparently not meant
+        to be used, hence we need our own. This is a quirk of subclassing
+        ndarray. See https://docs.scipy.org/doc/numpy/user/basics.subclassing.html
+        for more info.
+        """
```
jolyonb: Add a comment that `array_finalize` is unnecessary, as we don't add any properties to the object.

mitxgraders/helpers/calc.py
ChristopherChudzicki: I knew you were going to notice that change :).

On master:
```python
import numpy as np
from mitxgraders.helpers.calc import evaluator
result = evaluator('a/2 + a/2', {'a': np.array(8.0)}, {}, {})[0]
print(result) # prints 4.0
```
In-place operations here are problematic for mutable objects. It hadn't been a problem in earlier releases because parse_results only ever held immutable objects (float, int, complex, etc).

So, in fear of this issue I exorcised all in-place operations from calc.py.

I did add a test in `test_calc.py` to try and ensure that we don't reintroduce mutation later on.

jolyonb: Hmm, ok. I think `print(variables['a'])` was what you wanted to point me to here. I think we shouldn't be avoiding in-place operations. Instead, we should avoid exposing references to objects that we don't want mutated. In particular, the `eval_var` method (or whatever we called it) should return a copy of a variable, instead of a reference to the original.

jolyonb: An alternative is to use [augmented assignment](https://rszalski.github.io/magicmethods/) magic methods to return new objects instead of mutating current ones.

tests/helpers/calc/test_calc.py
```diff
@@ -144,3 +98,36 @@ def test_negation():
     for formula in badformulas:
         with raises(UnableToParse, match=re.escape(msg.format(formula))):
             evaluator(formula)
+
+def test_evaluation_does_not_mutate_variables():
+    """
+    This test should not be considered as related to vector/matrix/tensor algebra.
+
+    We're just trying to verify that variable variables aren't accidentally
```
jolyonb: Possible typo - variable variables? (mutable variables?)

mitxgraders/helpers/calc.py
ChristopherChudzicki: No, I did mean `print(results)`, but i actually didn't fully understand what was happening until you made me re-think it. Some logging explains:
```python
eval_product sees: [array(8.0), '/', 2.0]
variables is: {'a': array(8.0)}
eval_product sees: [array(4.0), '/', 2.0]
variables is: {'a': array(4.0)}
eval_sum sees: [array(2.0), '+', array(2.0)]
variables is: {'a': array(2.0)}
4.0
variables is: {'a': array(4.0)}
```

I like the idea of copying the variable values, that seems transparent. That would still require implementing __iadd__, etc MathArray, which is a bit of a pain, but I think mutative versions would actually be OK if we copied the variables.

mitxgraders/helpers/mathfunc.py
```diff
@@ -14,8 +14,80 @@
...
+    For now, scalar_domain() only accepts unary functions:
+    >>> @scalar_domain('add')
+    ... def f(x, y):
+    ...     return x + y
+    Traceback (most recent call last):
+    ValueError: Decorator 'scalar_domain' can only be used with unary functions.
+
+    Comment: The n-argument case seems intractable because it breaks get_number_of_args,
+    which is used in calc.py. But we could hard-code the n=1, n=2, n=3 cases.
+    Anything larger probably wouldn't come up in practice.
```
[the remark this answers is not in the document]
ChristopherChudzicki: If you use `*args`, then `get_number_of_args` does not correctly identify the number of arguments. for example

```pythong
def f(x, y, *args):
    return x + y + sum(args)

get_number_of_args(f) # returns 2
```

tests/helpers/calc/test_calc_arrays.py
```diff
@@ -0,0 +1,79 @@
...
+    A = MathArray([
+        [1, 5],
+        [4, -2]
+    ])
+    v = MathArray([3, -2])
+    n = 3
+    x = 4.2
+    z = 2 + 3j
+    variables = {
+        'A': A.copy(),
```
jolyonb: Why the copies?

mitxgraders/helpers/math_array.py
ChristopherChudzicki: Oh, good catch. Should be scalar. It's dot product.

mitxgraders/helpers/math_array.py
ChristopherChudzicki: Negative integers are fine if the matrix is numerically invertible.

Actually, that's worth thinking about more. Maybe there should be a matrix sampling class that ensures invertibility of the sample matrices.

tests/helpers/test_math_array.py
```diff
@@ -0,0 +1,461 @@
...
+    with raises(MathArrayError, match=match):
+        A + B
+
+    # dimension mismatch
+    u = MathArray([1, 2])
+    match = ('Cannot add/subtract a vector of length 2 with a matrix of '
+             'shape \(rows: 2, cols: 2\).')
+    with raises(MathArrayError, match=match):
+        u + B
+
```
jolyonb: Add test for two different length vectors.

mitxgraders/helpers/math_array.py
jolyonb: You already have vector * vector = number above...

mitxgraders/helpers/math_array.py
jolyonb: Waaait, numpy will do a matrix inversion automatically if you give it a negative power? Duuude!

I'm tempted to turn this off and maybe expose a function inv() to invert a matrix. Otherwise, matrix inversion problems will be trivially defeated, with no way to disallow the inverse operation. If inv is a function, the function can be blacklisted.

jolyonb: Also, there are so many matrix types to sample, we're going to have a field day writing samplers for them all ;-)

mitxgraders/helpers/calc.py
```diff
@@ -531,9 +531,19 @@ def eval_number(self, parse_result):
...
-            result = result * self.suffixes[parse_result[1]]
+            result *=  self.suffixes[parse_result[1]]
         return result
 
+    def eval_variable(self, parse_result):
+        """
+        Returns a copy of the variable in self.vars
+
+        NOTE: The variable's class must implement a __copy__ method.
+            (numpy ndarrays do implement this method)
```
jolyonb: Add a note for future benefit - this is to make sure that the variables in the dictionary don't mutate!

Also, does the copy method applied to a MathArray produce another MathArray?

Finally, does IdentityMatrix have a copy method?

mitxgraders/helpers/math_array.py
```diff
@@ -133,21 +152,24 @@ def __rsub__(self, other):
...
-                return super(MathArray, other).__mul__(self.item())
+                return super_MUL(other)
             elif other.ndim == 0:
-                return super(MathArray, self).__mul__(other.item())
+                return super_MUL(other.item())
             elif self.ndim > 2 or other.ndim > 2:
                 raise MathArrayError("Multiplication of tensor arrays is not currently supported.")
             try:
+                # can't mutate self during the multiplication
                 return np.dot(self, other)
```
jolyonb: Does this return a MathArray? I'm so confused with all the subclassing!

mitxgraders/helpers/math_array.py
ChristopherChudzicki: > I'm tempted to turn this off and maybe expose a function inv() to invert a matrix. Otherwise, matrix inversion problems will be trivially defeated, with no way to disallow the inverse operation. If inv is a function, the function can be blacklisted.

That's a good point. I have not made that change, but I did add two tests to assert that zero and negative powers work as (currently) expected, just so that I remember to document and test the behavior if we do change it.

mitxgraders/helpers/calc.py
jolyonb: I see that IdentityMatrix does indeed have copy.

mitxgraders/helpers/mathfunc.py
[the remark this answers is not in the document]
ChristopherChudzicki: I haven't changed anything, but I've tried to articulate the issue a little bit more in the docstring.

mitxgraders/helpers/mathfunc.py
```diff
@@ -168,13 +264,27 @@ def factorial(z):
...
-    # lambdas because sometimes np.real/imag returns an array,
-    're': lambda x: float(np.real(x)),
-    'im': lambda x: float(np.imag(x)),
+    'arccoth': arccoth
+}
+
+SCALAR_FUNCTIONS = {key: scalar_domain(key)(UNSAFE_SCALAR_FUNCTIONS[key])
+                    for key in UNSAFE_SCALAR_FUNCTIONS}
+
+ARRAY_FUNCTIONS = {
```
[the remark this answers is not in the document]
ChristopherChudzicki: Good point. That's fixed now.

mitxgraders/helpers/calc.py
ChristopherChudzicki: IdentityMultiple! =)

mitxgraders/helpers/mathfunc.py
```diff
@@ -66,26 +64,27 @@ def scalar_domain(display_name):
     Traceback (most recent call last):
     ValueError: Decorator 'scalar_domain' can only be used with unary functions.
 
-    Comment: The n-argument case seems intractable because it breaks get_number_of_args,
-    which is used in calc.py. But we could hard-code the n=1, n=2, n=3 cases.
-    Anything larger probably wouldn't come up in practice.
+    Comment: The n-argument case seems intractable because the _decorated_func
+    will be passed to get_number_of_args^[1], and so cannot have variadic signature *args.
+    But we could hard-code the n=1, n=2, n=3 cases.
+    [1]: In calc.py, eval_func calls is_callable_with, which uses get_number_of_args
```
jolyonb: Ah, ok, I see what's going on now. The decorator would work fine with *args, but get_number_of_args would later be called on the decorator, see only *args, and say "Zero!". Proposed solution: add a `nin` attribute to the decorator to specify the number of arguments that are expected.

ChristopherChudzicki: So I thought about this a bit more. 


In calc.py's `eval_function` action:
1. Currently, we use `is_callable_with_args` to check that every function will be called with the correct number of arguments. This has been a pain in the but because `is_callable_with_args` usings `get_number_of_args`, which is hard to make general.
2. But if we have decorated some functions with `@specify_domain`, then the decorated function knows how many arguments it should have and can validate that itself. This function would not need to be validated by `is_callable_with_args`. 

With that in mind, I implemented a version of `@specify_domain` using voluptuous. Here's the interface:

```python
@specify_domain(1, [2, 2], 2, display_name='puppy')
def f(x, y, z):
    return x*y*z

print(f.validated)
# True       specify_domain adds a .validated property to f, so that
#            eval_function knows that f does its own validation.

a = MathArray([1, 2])
b = MathArray([[1, 2, 3], [4, 5, 6]])
c = MathArray([1, 2])
f(a, b, c)
# DomainError: There was an error evaluating function puppy(...)
# Input 0 received a vector of length 2, expected a scalar
# Input 1 received a matrix of shape (rows: 2, cols: 3), expected matrix of shape (rows: 2, cols: 2)
# Input 2 is ok.
```

Thoughts? It would probably take me a few hours to write documentation and tests, so I wanted your thoughts first.

ChristopherChudzicki:
@jolyonb It seems to me that there is is a fair bit left to do: #71, #73 (easy), + Documentation and course examples. Do you want to merge this into a development branch? Alternatively, when you're ready to merge, I could remove `max_array_dim` configuration key, which essentially makes them unusable.

mitxgraders/helpers/mathfunc.py
jolyonb: I like the sound of this - it's much better than the hacky way we're using to validate functions, and we can expose the decorator too (or fall back to our hacky methods) for user functions.

jolyonb:
I've just been typing here - if you've been taking notes appropriately, I'll see if I can merge this one now.

jolyonb:
Ok, merging this for now. Further issues can be fixed in their own PRs.

PR #77, opened 2018-07-10
Specify domain

ChristopherChudzicki:
Adds class `SpecifyDomain`  (exported as `specify_domain`), an author-facing, configurable decorator that validates the inputs of the decorated function. For now, the only validation available is shape-validation.

Since SpecifyDomain is author-facing, its configuration is validated before it decorates a function. The unvalidated version `SpecifyDomain.make_decorator` is used internally to add input-validation to default functions like `sin` and `cos`.


mitxgraders/helpers/specify_domain.py
```diff
@@ -0,0 +1,241 @@
...
+def has_shape(shape):
+    if shape == (1,):
+        return number_validator
+    return make_shape_validator(shape)
+
+# how authors specify domain shape
+SCHEMA_SHAPE = Schema(Any(
+    All(Positive(int), lambda x: tuple([x])),
+    All([Positive(int)], Coerce(tuple)),
+    msg='Expected shape specification to be a positive integer or list of positive integers'
```
jolyonb: Can authors supply a numpy shape tuple?

mitxgraders/helpers/specify_domain.py
```diff
@@ -0,0 +1,241 @@
...
+    >>> b = MathArray([-1, 4])
+    >>> cross(a, b)                                 # doctest: +ELLIPSIS
+    Traceback (most recent call last):
+    DomainError: There was an error evaluating function cross(...)
+    ...1st input is ok: received a vector of length 3 as expected
+    ...2nd input has an error: received a vector of length 2, expected a vector of length 3
+    >>> cross(a)
+    Traceback (most recent call last):
+    DomainError: There was an error evaluating function cross(...): expected 2 inputs, but received 1.
+    """
```
jolyonb: Can you add a more complicated example here? I'm not sure from the example how to make a function that accepts a scalar and a matrix...

jolyonb: In particular, if input_shapes is a list, is that interpreted as a list of shape objects?

mitxgraders/helpers/specify_domain.py
```diff
@@ -0,0 +1,241 @@
...
+                for index, (shape, error) in enumerate(zip(shapes, errors)):
+                    ordinal = low_ordinal(index + 1)
+                    if error:
+                        lines.append('{0} input has an error: '.format(ordinal) + error.error_message)
+                    else:
+                        expected = 'scalar' if shape == (1,) else MathArray.get_description(shape)
+                        lines.append('{0} input is ok: received a {1} as expected'
+                            .format(ordinal, expected))
+
+                message = "\n\t".join(lines)
```
jolyonb: Tab characters look fine in bash, but they don't render in HTML. Do you plan to modify log output to transform tab characters into something meaningful?

mitxgraders/helpers/__init__.py
```diff
@@ -0,0 +1,6 @@
+from mitxgraders.helpers.specify_domain import specify_domain
+
+# Set the objects to be imported from this grader
```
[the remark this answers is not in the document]
ChristopherChudzicki: fixed

mitxgraders/helpers/specify_domain.py
ChristopherChudzicki: Not currently, and I think we should not allow it, though I am somewhat torn.

- numpy uses tuple for storing shape (I suspect because they don't want you mutating the shape property of an array). So it would be logical to use tuples for shape.
- On the other hand, in our configuration schemas, we uses lists for *almost* everything that requires enumeration. The only exception I can think of is using a tuple for enumerating multiple answers.

  In particular, suppose we use tuple for shape. We should be consistent, so then matrix sampling sets become:
  ```python
  RealMatrices(shape=(3, 2), norm=[1, 5])
  ```
  What the heck? Why is one a tuple and one is a list? That seems confusingly arbitrary!
- We could allow specifying shape as a list *or* tuple. But that's a pain from voluptuous's perspective, and makes documentation (a little bit) awkward.

mitxgraders/helpers/specify_domain.py
ChristopherChudzicki: > In particular, if input_shapes is a list, is that interpreted as a list of shape objects?
Yes, example added.

mitxgraders/helpers/specify_domain.py
ChristopherChudzicki: Good point. I just removed them, it made the doc tests fragile anyway.

jolyonb:
:+1:

PR #85, opened 2018-07-14
Consolidating error checking for NaN and inf

### LINKED ISSUE #67: Overflow errors not being caught
Apparently we have some issues with overflows not being caught in the equality comparer.

```python
from mitxgraders import FormulaGrader

grader = FormulaGrader(
    answers="1",
    debug=True
)

# These two lines raise two separate error messages, neither of which is processed into something useful for students
grader(None, "fact(100)")
grader(None, "fact(200)")
```

I also suspect that our factorial error messages are never being triggered any more, either.

jolyonb:
Resolves #67.

Consolidates error handling in calc.py and streamlines handling of infinities and NaNs.

### DIFF
--- a/mitxgraders/helpers/mitmath/calc.py
+++ b/mitxgraders/helpers/mitmath/calc.py
@@ -38,7 +38,7 @@ from mitxgraders.helpers.mitmath.mathfuncs import (
     DEFAULT_FUNCTIONS,
     DEFAULT_SUFFIXES
 )
-from mitxgraders.helpers.mitmath.math_array import MathArray
+from mitxgraders.helpers.mitmath.math_array import MathArray, IdentityMultiple
 from mitxgraders.helpers.mitmath.robust_pow import robust_pow
 from mitxgraders.helpers.mitmath.exceptions import (
     CalcError,
@@ -121,7 +121,7 @@ class BracketValidator(object):
             if bracket.is_closer:
                 try:
                     previous = stack.pop()
-                except IndexError: # happens if stack is empty
+                except IndexError:  # happens if stack is empty
                     BV.raise_close_without_open(formula, current)
                 if bracket.partner != previous.bracket.char:
                     BV.raise_wrong_closing_bracket(formula, current, previous)
@@ -253,7 +253,8 @@ def evaluator(formula,
               variables=DEFAULT_VARIABLES,
               functions=DEFAULT_FUNCTIONS,
               suffixes=DEFAULT_SUFFIXES,
-              max_array_dim=0):
+              max_array_dim=0,
+              allow_inf=False):
     """
     Evaluate an expression; that is, take a string of math and return a float.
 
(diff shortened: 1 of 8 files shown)


mitxgraders/helpers/mitmath/calc.py
```diff
@@ -573,10 +574,39 @@ def handle_node(node):
...
+                return float('nan')
+
+            # Compute the result of this node
+            result = action(handled_kids)
+
+            # All actions convert the input to a number, array or IdentityMultiple
+            # Check if there were any infinities or nan
+            check = result
+            # IdentityMultiple needs to extract the value
+            if isinstance(result, IdentityMultiple):
```
ChristopherChudzicki: Currently, `IdentityMultiple(0)` returns integer `0`. You could just make `identityMultiple(float('nan'))` return `nan`. (It seems like that approach would work for other custom algebraic objects too, like dimensions if people were ever interested in that again.)

jolyonb: What about `inf`?

mitxgraders/helpers/mitmath/calc.py
```diff
@@ -573,10 +574,39 @@ def handle_node(node):
...
+                check = result.value
+            if not allow_inf and np.any(np.isinf(check)):
```
ChristopherChudzicki: Currently our factorial function uses `math.factorial` if the input is an integer (or a float that can be interpreted as an integer) and `math.factorial` returns type `<long>` if its input is greater than 20. And `isinf` doesn't work for `<long>`.

Suggestion: scrap `math.factorial` all together, and go straight to the scipy gamma+1 implementation.
- The only difference will be `fact(3)=6.0` instead of `6`. I don't think this matters for anything at all, since MathArrays can handle integer-like powers.
- It would also resolve the issue of `fact(1000000)` timing out edX.

mitxgraders/helpers/mitmath/calc.py
jolyonb: I need to extract `inf` in this manner anyway, I think, because otherwise we'll have floats adding to matrices, no?

jolyonb: It would be ok if `inf` aborted a calculation, but we're not doing that...

mitxgraders/helpers/mitmath/calc.py
jolyonb: Ok, will do.

mitxgraders/helpers/mitmath/calc.py
ChristopherChudzicki: `IdentityMultiple(inf))` would need to stay `inf`, otherwise `1 + IdentityMultiple(inf)` would be scalar `inf`, which would be wrong.

ChristopherChudzicki: Oops. I understand. You extracted because isinf can't handle `IdentityMultiple`. Just ignore me :)

mitxgraders/helpers/mitmath/calc.py
jolyonb: Done.

mitxgraders/helpers/mitmath/calc.py
jolyonb: I updated `IdentityMultiple` for `nan`, but I think I still need to do this check to catch `inf`.

Issue #76, opened 2018-07-10
Refactor abs function to work on vectors


jolyonb:
Another thing that came up in comments and I don't want to lose. I saw that `norm` was added that provides essentially the same functionality, but the `abs` thinking and notation is definitely standard.

ChristopherChudzicki:
By the way: This was intentional on my part, although I suspected you might request a change.

When I teach multivariable calculus and linear algebra, I try very hard to use `norm/length/magnitude` for vectors, and reserve absolute value for scalars. (Incidentally, some error messages use length to refer to number of components. I might change..)

Anyway, I am very happy to change `abs` to work with vectors, but I just want to double check! But I do think we should keep norm as an alias for it.

jolyonb:
Resolved by #88.

PR #88, opened 2018-07-17
Class: MatrixGrader

### LINKED ISSUE #71: Separate class for grading vectors/matrices
Related to #70 

I suggest creating a subclass of FormulaGrader for grading vectors/matrices:

```python
class ArrayFormulaGrader(FormulaGrader):
  # custom configuration schema
  # custom default functions and variables
  # shadows check_answer
```

Reasons to create a new ArrayFormulaGrader class for grading vectors/matrices:

- Have separate default functions and variables. 
  - `I` for Identity: ArrayFormulaGrader could have a default variable named `I` for the identity operator.

    If FormulaGrader has a default variable named `I`, it is going to be useless most of the time, AND if authors try to overwrite it, they'll get warning errors (which can be suppressed, but still annoying).

  - Conceptually, it also makes sense to only provide `transpose` and `adjoint` functions when dealing with matrices. However, naming conflict is very unlikely for these.

- Separate Default Configuration: Currently, vectors and arrays are currently enabled through a `max_array_dim` configuration key. This can be hard-coded to 0 (scalars only) for FormulaGrader, and configurable for ArrayFormulaGrader (with a default of 2, for matrices). *Not a big deal, but I find it conceptually nice.*

- Customized Error Handling: Currently, *whenever* something goes predictably wrong with matrices, students get an error NOT marked incorrect. Examples:

  - multiplying / adding matrices with a shape mismatch
  - raising non-square matrix to powers

  I think that raising errors in this situation makes sense as a default. However, I can easily imagine a linear algebra course where the concept being assessed is shape mismatch. In that case, it might make sense to mark student incorrect for shape mismatch rather than raise an error.

  ArrayFormulaGrader could have configuration keys for this, easily implemented by shadowing `FormulaGrader.check_answer` and catching the appropriate exceptions.

  Of course, we could do this directly in FormulaGrader, but that class is, in my opinion, a bit bloated already.

ChristopherChudzicki:
This does a couple of things:

- minimize global defaults like `DEFAULT_FUNCTIONS`. Each mathy grading class now has class attributes `default_functions`, `default_variables`, and `default_suffixes` that determine the defaults used by that class.
- Adds new MatrixGrader class. MatrixGrader has some extra default variables/functions, and a few extra configuration options.

Should resolve #71, #76


mitxgraders/helpers/mitmath/mathfuncs.py
```diff
@@ -236,16 +249,29 @@ def has_one_scalar_input(display_name):
...
-    'norm': np.linalg.norm
+    'conj': np.conj
+}
+
+def has_one_square_input(display_name):
+    return SpecifyDomain.make_decorator('square', display_name=display_name)
+
+ARRAY_ONLY_FUNCTIONS = {
+    'norm': np.linalg.norm,
+    'abs': np.linalg.norm,
```
jolyonb: Can we split out `abs` into a function that only works on scalars and vectors (column/row)? Taking the absolute value of a matrix should return an error.

mitxgraders/helpers/mitmath/specify_domain.py
```diff
@@ -95,12 +107,33 @@ def make_shape_validator(shape):
     >>> validate_vec4('cat')
     Traceback (most recent call last):
     Invalid: received a str, expected a vector of length 4
+
+    Instead of specifying a tuple shape, you can speicfy 'square' to demand
```
jolyonb: specify

mitxgraders/helpers/mitmath/specify_domain.py
```diff
@@ -95,12 +107,33 @@ def make_shape_validator(shape):
...
+    >>> validate_square(square2)
+    MathArray([[1, 2],
+           [3, 4]])
+    >>> validate_square(square3)
+    MathArray([[1, 2, 3],
+           [4, 5, 6],
+           [7, 8, 9]])
+    >>> validate_square(MathArray([1, 2, 3, 4]))
+    Traceback (most recent call last):
+    Invalid: received a vector of length 4, expected a square matrix
```
jolyonb: This looks like a good place to make a `scalar_or_vector` input to implement `abs`.

jolyonb:
Oh, one more thing - can we make a list containing the Pauli matrices for me, so that I can add them as constants to a problem easily?

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -0,0 +1,58 @@
...
+
+class MatrixGrader(FormulaGrader):
+    """
+    An extension of FormulaGrader with better support for grading expressions
+    with vectors and matrices.
+
+    Configuration options as per FormulaGrader, plus:
+        max_array_dim (int): Specify the maximum array dimension that the
+            expression parser will accept, defaults to 2 (allows vectors
+            and matrices).
```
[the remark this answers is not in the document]
ChristopherChudzicki: Yes, variables can contain higher dimensions. I'll add a note about that.

mitxgraders/helpers/mitmath/calc.py
```diff
@@ -307,6 +302,10 @@ def evaluator(formula,
     >>> evaluator("inf", variables={'inf': float('inf')}, allow_inf=True)[0]
     inf
     """
+    variables = {} if variables is None else variables
+    functions = {} if functions is None else functions
+    suffixes = {} if suffixes is None else suffixes
```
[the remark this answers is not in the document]
ChristopherChudzicki: > So, I set these to the defaults to simplify both testing and use of evaluator by authors. 
> Now to get `pi`, `e` and `sin`, we need to do a bunch more typing.

I'm ok changing this back, but some things to consider:
- (Opinion): it's nice to keep things isolated / self-contained.
- In practice, none of the graders are affected by this change because they all use local variables / functions / suffixes. 

  There are a few isolated tests of `evaluator` by itself, and this change did break one of them: that test relied on `factorial(500.50` to return `inf`. But that test was a bit hard to understand unless you happen to know that `factorial(500.5)` is inf, so I used this as an opportunity to refactor that.

**But:** If you find yourself using `evaluator` in isolation for random development / testing purposes, I can see keeping the defaults and am happy to put them back.

mitxgraders/helpers/mitmath/specify_domain.py
ChristopherChudzicki: After I got my spell-chcker in atom, I switched linters, and it warned me near the beginning of these strings :(

mitxgraders/helpers/mitmath/calc.py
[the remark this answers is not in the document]
jolyonb: I agree that in practice, this changes nothing in the implementation, and with your opinion that things should be self-contained. However, we are exposing evaluator to problem authors if they desire to use it for anything, and I do routinely find myself using evaluator in isolation for dev/testing ;-)

mitxgraders/helpers/mitmath/specify_domain.py
ChristopherChudzicki: I decided just to hard-code the requirement in a separate absolute value function so I could add a custom message suggesting norm.

ChristopherChudzicki:
@jolyonb I think this is ready for re-review.

mitxgraders/helpers/mitmath/mathfuncs.py
```diff
@@ -255,9 +257,24 @@ def has_one_scalar_input(display_name):
...
 
+def array_abs(obj):
+    """
+    Takes absolute value of numbers or vectors and suggests norm(...) instead
+    for matrix/tensors.
+
+    NOTE: The decision to limit abs(...) to scalars and vectors was motivated
+    by pedagogy not software.
+    """
+    if isinstance(obj, MathArray) and obj.ndim > 1:
```
jolyonb: Very nitpicky here, but what happens to row/column matrices? (Do we even call these 'vectors'?) I'm likely to merge regardless, but just putting this out there as you think on #89.

mitxgraders/helpers/mitmath/mathfuncs.py
```diff
@@ -284,6 +301,21 @@ def merge_dicts(*source_dicts):
     'm': 1e-3, 'u': 1e-6, 'n': 1e-9, 'p': 1e-12
 }
 
+pauli = {
```
jolyonb: Hmm. Not sure I want these variables called `x`, `y` and `z` -- too likely to clash with other variable names (is there a field that calls them x/y/z, actually? I've seen a variety of conventions, but not that). Can we do `sigma_x`, `sigma_y` and `sigma_z`? (Also happy to hear opinions.)

jolyonb:
Agreed on #89. Everything else looks good; just holding off on the Pauli nomenclature, but good to merge after that.

mitxgraders/helpers/mitmath/mathfuncs.py
ChristopherChudzicki: That was ... silly of me. 

I don't know of any fields that call them lowercase `x`, `y`, and `z`. Capital `X`, `Y`, and `Z` are the standard in quantum information. (Well, 6 years ago anyway.)

Changed to `sigma_x`, etc. A QI course can change their names if desired.

PR #91, opened 2018-07-18
if matharray product is 1 item, return a number

### LINKED ISSUE #89: Issue with row*column multiplication
MathArray multiplication behaves ... unexpectedly...for row/column vectors:

```
a = MathArray([1, 2, 3]) # row: shape(1, 3)
b = MathArray([[1], [2], [3]]) # column: shape (3, 1)

(a*b).size # (1, 1)... a matrix with 1 row and 1 column
(a*b)*b # raises a ShapeError since (1,1) * (3, 1) does not work.
```

Relevant: MathArray mostly uses `np.dot` for multiplication, the documentation for which is very good: https://docs.scipy.org/doc/numpy/reference/generated/numpy.dot.html (Note that the documentation is for v1.13, we use 1.6. But I don't think `dot` has changed. It references `matmul` as preferred over `dot`, which does not exist in 1.6)

Suggested fix: MathArray's `__mul__` method has some checks to see if self or other has `ndim=0`. These checks should actually be self or size has `size=0`.

Documentation for numpy.ndarray.size: 

https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.size.html

ChristopherChudzicki:
This resolves #89. In multiplication, if output array contains a single item, convert it to a number.

### DIFF
--- a/mitxgraders/helpers/calc/math_array.py
+++ b/mitxgraders/helpers/calc/math_array.py
@@ -175,6 +175,8 @@ class MathArray(np.ndarray):
         elif isinstance(other, MathArray):
             if self.shape == other.shape:
                 return super_ADD(other)
+            elif is_numberlike_zero_array(self):
+                return self.item() + other
 
             msg = ("Cannot add/subtract a {self.description} with a {other.description}.").format(
                 self=self, other=other)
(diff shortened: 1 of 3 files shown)


jolyonb:
Very happy with this. Just the one comment.

jolyonb:
Note merge conflicts on this now.

ChristopherChudzicki:
@jolyon don't merge this yet, I want to check something

jolyonb:
Ok, whenever you're ready :)

ChristopherChudzicki:
Now it's ready.

jolyonb:
LGTM. Thanks for catching the finicky bugs!

Issue #74, opened 2018-07-09
Matrix equivalence grading


jolyonb:
We should be able to check how many entries in a tensor expression are correct or not, and assign credit appropriately. At least, it should be an option. Another option that could be enabled is a message that returns which entries are incorrect.

ChristopherChudzicki:
Sure. I think the way to do this would be to expand `comparer_function` to be like a normal customrespnse check function, returning either a boolean or a dict `{'ok': ..., 'msg': ..., 'grade_decimal': ...}`.

jolyonb:
Hmm. I'll need to think on this one. I want to keep the `comparer_function` implementation as simple as possible, as it's intended for other people to be able to write these straightforwardly.

ChristopherChudzicki:
Re comment in #94:
> I think these last few changes have made implementing #74 much easier now, too.
Yes, I think this would be easier now that comparer_function can return either True/False/'partial' or a dictionary with message and partial credit.

I think the biggest impediment is that currently FormulaGrader stops evaluating as soon as the first comparison fails (well, really, when `n_failures` exceeds `n_failable_evals`).

This would also really only be useful for matrix-entry problems. (Entering `[1, 2, 3]`, etc) as opposed to problems where we only want matrices for non-commutative symbols.

jolyonb:
Remember that the comparison function compares the entire matrix at once.

Also, agreed, this is only relevant for matrix-entry problems. If you're just manipulating expressions with array-like variables, this isn't a useful feature.

jolyonb:
Resolved by #255

PR #94, opened 2018-07-21
Grading incompatible types

ChristopherChudzicki:
Provides configuration keys to customize what should happen when a student submits an answer to MatrixGrader that has the wrong shape.

### DIFF
--- a/mitxgraders/formulagrader/matrixgrader.py
+++ b/mitxgraders/formulagrader/matrixgrader.py
@@ -3,10 +3,12 @@ matrixgrader.py
 
 Defines a FormulaGrader subtype that handles matrices, too.
 """
+from numbers import Number
+from collections import namedtuple
 from mitxgraders.formulagrader.formulagrader import FormulaGrader
 from voluptuous import Required, Any
 from mitxgraders.helpers.validatorfuncs import NonNegative
-from mitxgraders.helpers.calc import IdentityMultiple, MathArray
+from mitxgraders.helpers.calc import IdentityMultiple, MathArray, within_tolerance
 from mitxgraders.helpers.calc.exceptions import (
     MathArrayShapeError as ShapeError)
 from mitxgraders.helpers.calc.mathfuncs import (
@@ -64,7 +66,10 @@ class MatrixGrader(FormulaGrader):
             Required('max_array_dim', default=1): NonNegative(int),
             Required('negative_powers', default=True): bool,
             Required('shape_errors', default=True): bool,
-            Required('answer_shape_mismatch', default={}): {
+            Required('answer_shape_mismatch', default={
+                'is_raised': True,
+                'msg_detail': 'type'
+            }): {
                 Required('is_raised', default=True): bool,
                 Required('msg_detail', default='type'): Any(None, 'type', 'shape')
             }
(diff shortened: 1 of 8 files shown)


ChristopherChudzicki:
@jolyonb This is ready for review. I will address any changes you think should happen to this, but other than that I probably won't do anything until July 28th. (In particular, let me know if you think different default settings would be better.)

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -59,3 +80,31 @@ def check_response(self, answer, student_input, **kwargs):
...
+            msg_detail = self.config['answer_shape_mismatch']['msg_detail']
+
+            if msg_detail is None:
+                msg = ''
+            else:
+                detailed = msg_detail == 'shape'
+                expected_shape = get_description(comparer_params[0], detailed=detailed)
+                received_shape = get_description(student_input, detailed=detailed)
+                if not detailed and expected_shape == received_shape:
+                    msg = ("Exepected answer to be a {0}, but input is a {1}."
```
jolyonb: Expected

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -59,3 +80,31 @@ def check_response(self, answer, student_input, **kwargs):
...
+                           "of incorrect shape.".format(expected_shape, received_shape))
```
jolyonb: Wait, did this just run on from the period directly?

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -59,3 +80,31 @@ def check_response(self, answer, student_input, **kwargs):
...
+                msg = ("Exepected answer to be a {0}, but input is a {1}."
```
jolyonb: Expected again.

jolyonb:
Apart from some spelling/grammar, the implementation in `default_equality_comparer` is perfectly fine. However, can we take the try/except out of `default_equality_comparer` and wrap it around the call to the comparer, so that other comparers benefit from the same error handling?

mitxgraders/formulagrader/matrixgrader.py
ChristopherChudzicki: Do you mean this?
```python
continued_string = ( "a"
                                   "b" )
continued_string == "ab" # True
```

But the period was a typing. Looking at this more closely, apparently I forgot `assert` in all of my tests! (except the error-raising ones). So that needs to be fixed.

jolyonb: First line: "... but input is a {1}." (ends in a period).
Second line: "of incorrect shape."

I think you want to change the period to a space?

ChristopherChudzicki: Yep, and it didn't break the tests because I forgot the asserts. =/

ChristopherChudzicki:
Regarding
> However, can we take the try/except out of default_equality_comparer and wrap it around the call to the comparer, so that other comparers benefit from the same error handling?

I think you're right that we need a more generic solution here, but I'm not sure that this will work in general. In order to make appropriate error messages, we need to know what shape the answer should be. In the first version of this PR, the expected shape was determined from `comparer_params[0]` which, for the default comparer, is the actual answer. But for a general comparer, the shape of comparer_params[0] might not be the shape of the answer. (For example, for a comparer like `is_eigenvector_of`.)

So instead I moved the shape validation to a separate function, and this function is provided to the comparer function as part of `utils`.

## Also:
The original version of this PR had a number of issues that I've hopefully now resolved.

- types: fixed
- tests: In original PR I forgot to assert the comparisons, so several failing tests were not revealed. These have been fixed. Fixing the failing tests required:
    - `comparer_function`s can now return either `True`, `False`, `'partial'`, or a dictionary. This lets comparer messages pass specific failure messages.

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -59,3 +85,78 @@ def check_response(self, answer, student_input, **kwargs):
...
+        Assumes comparer_params is just the single expected answer wrapped in a list.
+        """
+        expected_input =comparer_params[0]
+        try:
+            # in numpy, scalars have empty tuples as their shapes
+            shape = tuple() if isinstance(expected_input, Number) else expected_input.shape
+            utils.validate_shape(student_input, shape)
+        except ShapeError as err:
+            if self.config['answer_shape_mismatch']['is_raised']:
+                raise err
```
jolyonb: Keep this as `raise`; it preserves the stack trace.

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -59,3 +85,78 @@ def check_response(self, answer, student_input, **kwargs):
...
+                          within_tolerance=_within_tolerance,
+                          validate_shape=_validate_shape)
+
+    def default_equality_comparer(self, comparer_params, student_input, utils):
+        """
+        Default comparer function.
+
+        Assumes comparer_params is just the single expected answer wrapped in a list.
+        """
+        expected_input =comparer_params[0]
```
jolyonb: Linting nit: space after `=`!

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -59,3 +85,78 @@ def check_response(self, answer, student_input, **kwargs):
...
+        try:
+            # in numpy, scalars have empty tuples as their shapes
+            shape = tuple() if isinstance(expected_input, Number) else expected_input.shape
+            utils.validate_shape(student_input, shape)
+        except ShapeError as err:
```
jolyonb: Can we move this try/except block into MatrixGrader, so that comparers can just raise a ShapeError and it's treated appropriately? Open to suggestions here.

jolyonb:
Looking much better. Few very minor comments, one small refactor comment.

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -58,4 +90,78 @@ def check_response(self, answer, student_input, **kwargs):
                 raise
             else:
                 return {'ok': False, 'msg': err.message, 'grade_decimal': 0}
+        except InputTypeError as err:
```
ChristopherChudzicki: Re old comment:
> Can we move this try/except block into MatrixGrader, so that comparers can just raise a ShapeError and it's treated appropriately? Open to suggestions here.

I moved the try/except block out of the comparer, but in order to do what we want (if I understood correctly) we need a new error type. Reason is that ShapeErrors happen for two reasons:

- Student input contains shape errors like `"[[1, 2, 3]] - 2`
- Author encounters a shape error when subtracting `comparer_params[0] - student_input`. In this case it would be confusing to student to see "Cannot add/subtract matrix and scalar", since student doesn't "see" any subtraction happening. So these errors are re-raised as InputTypeErrors.

jolyonb:
LGTM 👍

jolyonb:
I think these last few changes have made implementing #74 much easier now, too.

PR #107, opened 2018-08-05
Improve isinf isnan checks

### LINKED ISSUE #104: handle_node breaks when functions have arguments of different types
A very minimal example:

```pycon
>>> import numpy as np
>>> np.isinf( np.array([1, 2, 3]), 5 )
TypeError: return arrays must be of ArrayType
```

I don't completely understand the issue, but apparently `np.isinf` cannot handle a list where some members are arrays and some members are not. Same issue with `np.isnan`.

This shows up for us in `calc.py`, where the `self.actions['arguments']` outputs a list of function arguments, some of which might be arrays, some of which might be numbers.

Here's a full example that produces the error:

```python
import numpy as np
from mitxgraders import MatrixGrader, RealVectors

def rot(vec, axis, angle):
    """
    Rotate vec by angle around axis. Implemented by Euler-Rodrigues formula:
    https://en.wikipedia.org/wiki/Euler%E2%80%93Rodrigues_formula
    """
    unit_axis = axis/np.linalg.norm(axis)
    a = np.cos(angle/2)
    omega = unit_axis * np.sin(angle/2)
    crossed = np.cross(omega, vec)

    return vec + 2*a*crossed + 2*np.cross(omega, crossed)


grader = MatrixGrader(
    answers='rot(a, [0, 0, 1], theta)',
    variables=['a', 'theta'],
    sample_from={
        'a': RealVectors(shape=3),
    },
    user_functions={
        'rot': rot
  },
  debug=True
)
```

(Discovered while writing documentation for `specify_domain`)

ChristopherChudzicki:
Resolves #104

### DIFF
--- a/tests/helpers/calc/test_calc.py
+++ b/tests/helpers/calc/test_calc.py
@@ -12,6 +12,7 @@ from mitxgraders.helpers.calc.exceptions import (
     UnbalancedBrackets, UndefinedVariable,
     ArgumentError, CalcOverflowError, CalcZeroDivisionError
 )
+from mitxgraders.helpers.calc.math_array import equal_as_arrays, MathArray
 
 def test_calcpy():
     """Tests of calc.py that aren't covered elsewhere"""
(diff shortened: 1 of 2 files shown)


jolyonb:
Can you rebase this please?

ChristopherChudzicki:
Rebased!

tests/helpers/calc/test_calc.py
```diff
@@ -103,13 +104,15 @@ def test_calc_functions_multiple_arguments():
...
 
-    def h3(x, y, z): return x + y + z
+    def h3(x, y, z): return x * y * z
 
     assert evaluator("h(2)", {}, {"h": h1}, {})[0] == 2.0
-    assert evaluator("h(1, 2)", {}, {"h": h2}, {})[0] == 3.0
-    assert evaluator("h(1, 2, 3)", {}, {"h": h3}, {})[0] == 6.0
+    assert evaluator("h(2, 3)", {}, {"h": h2}, {})[0] == 6.0
+    assert evaluator("h(2, 3, 4)", {}, {"h": h3}, {})[0] == 24.0
+    assert equal_as_arrays(evaluator("h(2, [1, 2, 3])", {}, {"h": h2})[0],
```
ChristopherChudzicki: By the way: Line 114 would have failed before this PR, but succeeds now. I changed the `h`s to be multiplication so that in `h2(x, y)`, the arguments `x` and `y` could be number and MathArray. (which would not work for addition).

Oh. Well I guess I could have used addition and `0`. Oh well.

jolyonb:
Ok, this looks fine. I finally figured out where on earth you ran into this issue... only the `arguments` handler ever returns a list, and the arguments handler has never before seen anything funky!

PR #106, opened 2018-08-05
Version 1.2 documentation

ChristopherChudzicki:
- [x] documentation for `MatrixGrader`
    - course examples:
      - [x] basic example illustrating algebra, non-commutative behavior, error messages
      - [x] matrix entry-by-entry input example
      - [x] cross product example
      - [x] example highlighting vector vs single-row-matrix vs single-column-matrix
- [x] documentation for vector/matrix sampling classes
- [x] `specify_domain` documentation
    - [x] course example
- [x] document `comparer_function`s
    - [x] `NumericalGrader` course example using `is_between` comparer
    - [x] `MatrixGrader` course example using eigenvector/eigenvalue comparer
    - [x] `ListGrader` eigenvalue/eigenvector comparer demo
- [x] documentation for sibling variables when FormulaGrader used in ListGrader
    - [x]  course examples
- [x] consolidate `index.md` and `getting_starting.md` (there was a lot of overlap)

Some of these are very close to done.

[My version of the documentation website](https://christopherchudzicki.github.io/mitx-grading-library/)
[My version of demo course](https://edge.edx.org/courses/course-v1:CAC+CAC103+FOREVER/course/)

Note: I changed the theme to [mkdocs-cinder](https://github.com/chrissimpkins/cinder), because I wanted one level deeper on the individual page table of contents. Could explore other themes, but not going to worry about it now.

PR #109, opened 2018-08-05
add a numerical hard_tolerance to within_tolerance

ChristopherChudzicki:
This PR resolves the following issue:

```python
grader = FormulaGrader(answers='0')
grader(None, 'sin(pi)')['ok']      # False, due to rounding errors
                                   # and grader's default tolerance of 0.01%
```

by changing `within_tolerance` to accept anything within `hard_tolerance` of the answer (default: `1e-12`).

Before this PR, authors could work around this issue by setting a numerical tolerance for the grader.  But consider

-  My opinion is that we shouldn't make people think about.
-  Although the issue is somewhat transparent in the simple example above, it is a bit harder to understand in some situations. (I came across this while implementing `eigenvector_comparer`, https://github.com/mitodl/mitx-grading-library/pull/106/files#diff-8d2928c7449112fa868ac7f32db680d3R179)

### DIFF
--- a/mitxgraders/helpers/calc/mathfuncs.py
+++ b/mitxgraders/helpers/calc/mathfuncs.py
@@ -316,14 +316,17 @@ pauli = {
     ])
 }
 
-def within_tolerance(x, y, tolerance):
+def within_tolerance(x, y, tolerance, hard_tolerance=1e-12):
     """
-    Check that |x-y| <= tolerance with appropriate norm.
+    Check that |x-y| <= tolerance OR hard_tolerance with appropriate norm.
 
     Args:
         x: number or array (np array_like)
         y: number or array (np array_like)
         tolerance: Number or PercentageString
+        hard_tolerance: Number
+            hard_tolerance is intended to combat rounding errors. See usage
+            below.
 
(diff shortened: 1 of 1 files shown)


jolyonb:
Note that there are two types of tolerances: absolute and relative. Your hard cutoff is setting an absolute tolerance. The default is a relative tolerance, because that's typically a better fit.

Your hard cutoff fix isn't going to work everywhere. It might be ok for formula graders, but for numerical graders, it's a disaster if the answer is 10^-13 or smaller, because you've just mandated an absolute tolerance that is orders of magnitude bigger than the answer.

ChristopherChudzicki:
>  but for numerical graders, it's a disaster if the answer is 10^-13 or smaller,

True.

Honestly, I probably need a better understanding of floating point numbers. I thought that since 

```pycon
>>> 1.12345678910111213141516 == 1.12345678910111213141517
True
```
it would be the case that

```pycon
>>> 0.0 == 1e-20
True # This is false
```

**Question: **  Do you think this issue needs a better resolution? Or is the resolution just to use absolute tolerances if you want `'0'` and `np.sin(np.pi)` to be the same?

jolyonb:
We may like to document this better though!

Issue #110, opened 2018-08-06
Rename Trace


jolyonb:
Instead of `tr`, require `trace`. It can render as `Tr` though.

ChristopherChudzicki:
Resolved by #111

PR #111, opened 2018-08-06
Renaming tr to trace

jolyonb:
Quick fix to rename `tr` to `trace`, so as not to confuse with `transpose`. Also added `trace` to javascript file. Documentation not updated, as this is part of a separate PR.


jolyonb:
Resolves #110.

mitxgraders/helpers/calc/mathfuncs.py
```diff
@@ -277,7 +277,7 @@ def array_abs(obj):
     'abs': array_abs,
     'trans': np.transpose,
     'det': has_one_square_input('det')(np.linalg.det),
-    'tr': has_one_square_input('tr')(np.trace),
+    'trace': has_one_square_input('tr')(np.trace),
```
ChristopherChudzicki: Let's use `trace` for the display name, too.

Issue #95, opened 2018-07-22
Functions to implement in asciimath javascript


jolyonb:
These are the new functions that we've introduced that don't have a corresponding function in AsciiMath:

* `trace`: needs AsciiMath definition (done in #111)
* `trans`, `ctrans`, `adj`, `cross`: needs preprocessing
* `conj`: needs an option to write as `(...)^*`
* Vectors need an option to display as column vectors

jolyonb:
Here's some of the replacements we want:

* `conj(...)` -> `{:(...)^*:}` with option set
* `trans(...)` -> `{:(...)^T:}`
* `adj(...)` -> `{:(...)^dagger:}`
* `ctrans(...)` -> `{:(...)^dagger:}`
* `cross(..., ...)` -> `(...) times (...)`
* `[a,b,c]` -> `[[a],[b],[c]]` with option set if a, b, and c don't have square brackets in them
* `fact(...)` and `factorial(...)` -> `factorialAsciiMath((...))`

Here's the dagger symbol definition:
````javascript
// Add the dagger symbol, used in Hermitian conjugate/adjoint
AM.newsymbol({input:"dagger", tag:"mi", output:"\u2020", tex:null, ttype:AM.TOKEN.CONST});
````

In all cases, single character entries should not have parentheses. That includes all entries in this list:
````
['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 'alpha', 'beta', 'chi', 'delta', 'Delta', 'epsi', 'varepsilon', 'eta', 'gamma', 'Gamma', 'iota', 'kappa', 'lambda', 'Lambda', 'lamda', 'Lamda', 'mu', 'nu', 'omega', 'Omega', 'phi', 'varphi', 'Phi', 'pi', 'Pi', 'psi', 'Psi', 'rho', 'sigma', 'Sigma', 'tau', 'theta', 'vartheta', 'Theta', 'upsilon', 'xi', 'Xi', 'zeta']
````
[...]

PR #116, opened 2018-08-07
Javascript updates

jolyonb:
This PR updates the javascript preprocessor substantially.

* `dagger` is defined in AsciiMath
* `fact` and `factorial` are fixed
* `ctrans` and `adj` are implemented
* `replaceFunctionCalls` has some extra logic added to make sure that it only affects the desired functions
* `conj` is implemented as a star when the option is set
* Lots of tests!

### DIFF
--- a/MJxPrep.js
+++ b/MJxPrep.js
@@ -41,6 +41,10 @@ if (window.MJxPrep) {
     // log2(x) -> log_2(x)
     eqn = eqn.replace(/log2\(/g, "log_2(");
 
+    // Match Deltaxyz and deltaxyz and wrap in invisible brackets for display
+    // So, DeltaE -> {:DeltaE:}, deltasomething -> {:deltasomething:}
+    eqn = eqn.replace(/([Dd]elta)([a-zA-Z]+)/g, "{:$1$2:}");
+
     // Factorial: We want fact(n) -> n!, but fact(2n) -> (2n)!
     // Replace fact(...) -> {:factAsciiMath((...)):}, with inner parentheses added as necessary
     eqn = replaceFunctionCalls(eqn, 'fact', function(funcName, args) {
--- a/MJxPrep.js
+++ b/MJxPrep.js
@@ -12,12 +12,17 @@
 // Make sure that this script only loads once
 if (window.MJxPrep) {
 } else {
+  // Specify options
+  window.MJxPrepOptions = {
+    'conj_as_star': true
+  }
+
   // Define the preprocessor
   window.MJxPrep = function() {
-    /*-----------------------------------------------------
-         *This is the preprocessor, used for translating inputs
-         *-----------------------------------------------------
-         */
+    /*------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jolyonb:
This partially addresses #95. Still more to be done there, however.

jolyonb:
This now also wraps Deltax and deltax appropriately.

MJxPrep.js
```diff
@@ -270,11 +296,26 @@ if (window.MJxPrep) {
     // default value for startingAt
     var startingAt = startingAt ? startingAt : 0
 
-    // Replace the first instance of 'funcName(<args>)'
-    var funcStart = expr.indexOf(funcName, startingAt)
-    if (funcStart < 0) {
-      return expr
-    }
+    // Find the first instance of 'funcName(<args>)' we care about
+    do {
```
ChristopherChudzicki: Good idea to check for `(` and previous alphabetic characters.

**Opinion:** When reading this, I found the `do...while...` loop hard to reason about, so it took me a bit to understand what was happening. (in particular, `while(true)` especially looks scary, but I found the `break` 😄)

Suggestion: We're already doing this recursively, so how would you feel about replacing the `do` loop with something like:

```js
  function replaceFunctionCalls(expr, funcName, action, startingAt) {
    // default value for startingAt
    var startingAt = startingAt ? startingAt : 0
    var funcStart = expr.indexOf(funcName + '(', startingAt);

    if (funcStart < 0) { return expr; }
    // Make sure the previous character isn't an alpha character
    // (don't match the end of a function name we don't want to match)
    // This will allow us to replace "f(...)" without replacing "diff(...)"
    if( funcStart > 0 && /[a-zA-Z]/.test(expr.substr(funcStart-1, 1)) ) {
      // False positive, look further
      return replaceFunctionCalls(expr, funcName, action, funcStart + 1)
    }

  // etc ...
  }
```

MJxPrep.js
```diff
@@ -12,12 +12,17 @@
 // Make sure that this script only loads once
 if (window.MJxPrep) {
 } else {
+  // Specify options
+  window.MJxPrepOptions = {
+    'conj_as_star': true
```
ChristopherChudzicki: Just checking: We are expecting people to ...

1. decide on course-wide options for these things
2. download the js file
3. customize the js file
4. include at appropriate places in their course

Is that correct? I was thinking about this, and that's the best I could come up with.

jolyonb: Pretty much. Note that including at appropriate locations typically means loading it as a preprocessor for a textbox. Without edX making anything useful, I think this is the best we've got.

MJxPrep.js
jolyonb: Sure, can do that.

jolyonb:
Updated and ready for review.

ChristopherChudzicki:
> This now also wraps Deltax and deltax appropriately.
👍
Looks good to me, so merging it.

Issue #117, opened 2018-08-07
MJxPrep needs better error handling


ChristopherChudzicki:
Consider the following typing:

```
                        // displays as:
cross(a,b) + fact       // a x b + fact
cross(a,b) + fact(      // cross(a,b) + fact(
cross(a,b) + fact(5     // cross(a,b) + fact(5
cross(a,b) + fact(5)    // a x b + 5!
```

errors in one part of a string should not prevent display of the rest.

I have a plan, but finishing column vector display first.

jolyonb:
Yeah, I saw this. It could definitely be improved, but it's a bit trickier than I want to deal with in javascript! All yours :-)

PR #118, opened 2018-08-07
Js updates 2

ChristopherChudzicki:
Display cross products and vectors as described in #95

### DIFF
--- a/MJxPrep.js
+++ b/MJxPrep.js
@@ -50,12 +50,12 @@ if (window.MJxPrep) {
     // Replace fact(...) -> {:factAsciiMath((...)):}, with inner parentheses added as necessary
     eqn = replaceFunctionCalls(eqn, 'fact', function(funcName, args) {
       validateArgsLength(funcName, args, 1)
-      return '{:factAsciiMath(' + groupExpr(args[0]) + '):}'
+      return '{:' + groupExpr(args[0]) + '!:}'
     } )
     // Replace factorial(...) -> {:factAsciiMath((...)):}, with inner parentheses added as necessary
     eqn = replaceFunctionCalls(eqn, 'factorial', function(funcName, args) {
       validateArgsLength(funcName, args, 1)
-      return '{:factAsciiMath(' + groupExpr(args[0]) + '):}'
+      return '{:' + groupExpr(args[0]) + '!:}'
     } )
 
     // Transpose: trans(x) -> x^T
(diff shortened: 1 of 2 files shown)


MJxPrep.js
```diff
@@ -317,17 +366,66 @@ if (window.MJxPrep) {
...
+    // function calls
+    var processedArgsString = replaceFunctionCalls(argsString, funcName, action)
+    var args = shallowListSplit(processedArgsString)
+
+    // perform the action
+    var finished = expr.substring(0, funcStart) + action(funcName, args)
+    var unfinished = expr.substring(closeCallParens + 1)
+
+    // Recursively process the unfinished string
+    return replaceFunctionCalls(finished + unfinished, funcName, action, finished.length + 1)
```
ChristopherChudzicki: Earlier implementation of `replaceFunctionCalls` could be made to enter an infinite loop (see below). This fixes that.

```js
            //0123456890
const expr = 'cat(1) + 1'
const action = (funcName, args) => 'gonna crash cat(1)'
replaceFunctionCalls(expr, 'cat', action)
```
The original expression closes at position 5. When `replaceFunctionCalls` calls itself the first time, it starts searching at postion 5+1, but that 'cat' has been pushed past that position, so it is encountered again ...

MJxPrep.js
```diff
@@ -50,12 +50,12 @@ if (window.MJxPrep) {
     // Replace fact(...) -> {:factAsciiMath((...)):}, with inner parentheses added as necessary
```
jolyonb: Comment needs fixing now.

MJxPrep.js
```diff
@@ -50,12 +50,12 @@ if (window.MJxPrep) {
...
     eqn = replaceFunctionCalls(eqn, 'fact', function(funcName, args) {
       validateArgsLength(funcName, args, 1)
-      return '{:factAsciiMath(' + groupExpr(args[0]) + '):}'
+      return '{:' + groupExpr(args[0]) + '!:}'
     } )
     // Replace factorial(...) -> {:factAsciiMath((...)):}, with inner parentheses added as necessary
```
jolyonb: And here too.

jolyonb:
Sorry for being picky!

jolyonb:
Does this complete our todo list on javascript?

ChristopherChudzicki:
> Does this complete our todo list on javascript?

I think it completes #95, but I'm going to go resolve #117

PR #119, opened 2018-08-07
Adding some extras

jolyonb:
This adds some brackets (important for bras and kets!) as well as hbar to the definition list. None of these are going to interfere with anybody, but this definitely makes my life a little easier... I also added some comments.


jolyonb:
@ChristopherChudzicki Can you review this briefly?

jolyonb:
Sweet, thanks.

PR #120, opened 2018-08-07
process string even if some errors present

### LINKED ISSUE #117: MJxPrep needs better error handling

ChristopherChudzicki:
Resolves #117.

### DIFF
--- a/MJxPrep.js
+++ b/MJxPrep.js
@@ -244,10 +244,18 @@ if (window.MJxPrep) {
   // Check for the AsciiMath object every 200ms
   var checkExist = setInterval(updateMathJax, 200);
 
-  function findClosingBrace(expr, startIdx) {
+  /**
+   * get index at which the brace at braceIdx in expr is closed, or null if it
+   * does not close.
+   *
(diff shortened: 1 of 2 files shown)


jolyonb:
We also need to deal with `validateArgsLength`, which raises an error if there's an argument problem, rather than just saying "Skip this operation".

ChristopherChudzicki:
> We also need to deal with validateArgsLength, which raises an error if there's an argument problem, rather than just saying "Skip this operation".

Good catch

ChristopherChudzicki:
Should be good now.

jolyonb:
Looks good. I'll merge once tests are done. Thanks for this!

PR #123, opened 2018-08-08
disallow multiplication of 3+ vectors w/o parens

### LINKED ISSUE #108: [Problematic] Vector multiplication is not associative
Since `*` multiplies vectors as a dot product, `*` is not associative when used with vectors.

```python
from mitxgraders.helpers.calc import evaluator, MathArray

variables = {
    'i': MathArray([1, 0]),
    'j': MathArray([0, 1]),
}

evaluator("i*i*j", variables)[0]        # result: [0, 1, 0]
evaluator("i*(i*j)", variables)[0]      # result: [0, 0, 0]
```

This seems like it could cause confusion, especially since edX renders multiplication with dot product symbols.

**Edit**: In particular, students might not know above whether `i*i*j` is equivalent to `(i*i)*j` or `i*(i*j)`.

ChristopherChudzicki:
Resolves #108 

Error message is potentially confusing in large expressions. Would be great if `parse_results` included the original string so we could show students what string caused the error. One day, maybe... !

### DIFF
--- a/mitxgraders/helpers/calc/calc.py
+++ b/mitxgraders/helpers/calc/calc.py
@@ -958,6 +958,12 @@ class FormulaParser(object):
         Traceback (most recent call last):
         CalcError: Unexpected symbol + in eval_product
         """
+        num_vectors = sum([isinstance(operand, MathArray) and operand.ndim == 1
+                           for operand in parse_result])
+        if num_vectors >= 3:
+            raise CalcError("Multiplying three or more vectors is ambiguous. "
+                            "Please place parentheses around vector multiplications.")
+
         result = parse_result[0]
         data = parse_result[1:]
         while data:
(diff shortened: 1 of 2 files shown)


jolyonb:
Looks fine to me.

jolyonb:
Ah. I've had second thoughts... We shouldn't give this message for `vecv/2/pi`...

ChristopherChudzicki:
>

We don't, but I think I see what you mean. Currently:

```python
from mitxgraders.helpers.calc import evaluator, MathArray

variables = {
    'i': MathArray([1, 0]),
    'j': MathArray([0, 1]),
}

result = evaluator("i/i*i", variables)[0]
```

raises an error, but it should just give `MathArray([1, 0, 0])`

PR #126, opened 2018-08-09
Convert np scalars

### LINKED ISSUE #124: MathArrays behave badly with numpy's numeric types
Behold, my nightmare:

```python
from mitxgraders.helpers.calc import evaluator
variables = {'v': MathArray([3, 4])}
evaluator('v*v + v') # scalar + vector,
                     # but does not throw error
```

## Explanation:

```python
import numpy as np
from mitxgraders.helpers.calc import MathArray

x = MathArray([1, 2, 3])

1.0 + x                 # throws an error, good!
np.float64(1.0) + x     # returns MathArray([2.0, 3.0, 4.0])
```

Here's what's happening:

- `1.0 + x`:
    - tries to call `(1.0).__add__(self, x)` but fails, because builtin float type does not know how to handle `MathArray`
    - next tries `x.__radd__(self, 1.0)`, which knows that it should through a StudentFacingError.
- `np.float64(1.0) + x`:
    - tries to call `np.float64(1.0).__add__(self, x)` and succeeds: Numpy's `float64` numeric type knows how to add `ndarray`s to itself.
    - I.e., this version never calls any of `MathArray`'s methods

I'm guessing this is true for all numeric operators and true for all numpy numeric types. Bummer.

Note: `np.float64.__add__` is implemented in C. So even if we wanted to monkey-patch it, I don't think we could.

## Goals:

  - Resolve this issue (insofar as it could ever affect students) 
  - continue to keep most of the type/shape checking outside of calc.py

ChristopherChudzicki:
Resolves #124 

Also implements #123 without errors.

### DIFF
--- a/mitxgraders/helpers/calc/calc.py
+++ b/mitxgraders/helpers/calc/calc.py
@@ -342,6 +342,36 @@ def evaluator(formula,
                        suffixes=math_interpreter.suffixes_used)
     return result, usage
 
+def cast_np_numeric_as_builtin(obj):
+    """
+    Cast numpy numeric types as builtin python types.
+
+    NOTE: We do this because instances of np.number have their own __radd__
+    method for handling arrays, which circumvents MathArray's type/shape
+    checking.
+
+    Examples:
+    >>> import numpy as np
+    >>> x = 1.0
+    >>> x64 = np.float64(x)
+    >>> y = 5
+    >>> y64 = np.int64(y)
+    >>> z = 3 + 2j
+    >>> z128 = np.complex128(z)
+    >>> examples = [x, x64, y, y64, z, z128]
+    >>> [type(cast_np_numeric_as_builtin(example)) for example in examples]
+    [<type 'float'>, <type 'float'>, <type 'int'>, <type 'int'>, <type 'complex'>, <type 'complex'>]
+
+    Leaves MathArrays alone:
+    >>> from mitxgraders.helpers.calc.math_array import MathArray
+    >>> A = MathArray([1, 2, 3])
+    >>> cast_np_numeric_as_builtin(A)
+    MathArray([1, 2, 3])
+    """
+    if isinstance(obj, np.number):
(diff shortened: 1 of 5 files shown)


mitxgraders/helpers/calc/calc.py
```diff
@@ -591,7 +620,7 @@ def handle_node(node):
             if any(np.any(np.isnan(r)) for r in as_list):
                 return float('nan')
 
-            return result
+            return cast_np_numeric_as_builtin(result)
```
jolyonb: Shouldn't this be mapped across lists?

jolyonb:
Looks good. Just the one comment, I think.

mitxgraders/helpers/calc/calc.py
ChristopherChudzicki: I agree—good idea—and it's done. That said, I couldn't find a place where it mattered. (`result` should only be a list if it is the result of `eval_arguments`, and in that case all the list items should already have passed through `handle_node`). Anyway, I agree, but that's why there's no new test asserting this behavior.

mitxgraders/helpers/calc/calc.py
```diff
@@ -591,7 +630,7 @@ def handle_node(node):
             if any(np.any(np.isnan(r)) for r in as_list):
                 return float('nan')
 
-            return result
+            return cast_np_numeric_as_builtin(result, map_across_lists=False)
```
jolyonb: Did you mean to set this True?

mitxgraders/helpers/calc/calc.py
jolyonb: Agreed, I'm not sure it matters presently, but future-proofing!

mitxgraders/helpers/calc/calc.py
ChristopherChudzicki: Oops, that was a relic from trying to find a test that would break when `False`. Fixed.

PR #129, opened 2018-08-12
Updating docs

jolyonb:
And some more doc fixes (mostly trouble with nested lists). I had trouble running `mkdocs gh-deploy`, because the `cinder` theme for `mkdocs` isn't compatible with the latest version of `mkdocs`. We should probably look into a new theme at some point. I had to restrict `mkdocs` to an earlier version.


jolyonb:
@ChristopherChudzicki I have succeeded in pushing the documentation from this branch!

jolyonb:
Note, I tried to put a `README.md` file in the `gh-pages` branch, but it got deleted upon a subsequent update.

Issue #132, opened 2018-09-13
Suffix match order bug


ChristopherChudzicki:
Note: I came across this bug while rewriting the parser/evaluator. I'm creating the issue mostly so I can reference it in a future PR. (A PR which would also fix this bug.)

```python
from mitxgraders.helpers.calc import evaluator

variables = {}
functions = {}
suffixes_ok = {  'ab': 2 }
suffixes_error = { 'a': 10, 'ab': 2 }

expr = '5ab'

evaluator(expr, variables, functions, suffixes_ok) # works as expected
evaluator(expr, variables, functions, suffixes_error) # throws an error
```

The issue is that `a` is a front substring of `ab`. I think that the grammar was originally intended to handle this case because in defining suffixes, `MatchFirst` is used, which can be used to specify the order in which to check. (So above, `MatchFirst(['ab', 'a'])` would have been fine.) But we give `MatchFirst` a bunch of dictionary keys, which aren't ordered:

```python
# Define our suffixes
suffix = MatchFirst(Literal(k) for k in self.suffixes.keys())
```

ChristopherChudzicki:
Incidentally, it is also possible to do some pretty silly things with suffixes:

```python
import numpy as np
from mitxgraders.helpers.calc import evaluator

variables = {}
functions = {}
suffixes_ok = { '-1': 10 }

expr = '5-1'

result, _ = evaluator(expr, variables, functions, suffixes_ok)
print(result) # 10.0
```

jolyonb:
You sure that doesn't give you 50?

I was aware the suffixes had these issues, but since we were restricting them to % and hard-coded metric suffixes, it seemed safe enough. Feel free to fix if you like. I suspect that it should probably restrict to single character suffixes that cannot include certain characters.

jolyonb:
Resolved by #134

PR #134, opened 2018-09-16
rewrite expression parser/evaluator

ChristopherChudzicki:
Rewrote much of `calc.py` and changed name to `expressions.py`

Noteable changes:

1. I tried to separate parsing and evaluation as much as possible. Usage now looks like:

    ```python
    parser = MathParser()
    expr_1 = parser.parse('x + y')
    expr_2 = parser.parse('x*y')
    
    variables_A = { 'x': 5, 'y': 7 }
    variables_B = {'x': -3, 'y': 4}
    
    expr_1.eval(variables_A, {}, {}, {})
    expr_1.eval(variables_B, {}, {}, {}) # evaluate the same expression with different variables, do not re-parse
    
    expr_2.eval(variables_A, {}, {}, {})
    expr_2.eval(variables_B, {}, {}, {}) # evaluate the same expression with different variables, do not re-parse
    ```
2. I changed the grammar so that the **parser** accepts all alphabetic suffixes and '%'. Advantages:
    - Allows us to give better error messages for things like `'x + 5y'`. We used to just say "Unable to parse". Now the expression DOES parse, but we can say "y not allowed directly after number, did you forget `*`?"
    - Resolves the weirdness in #132

### DIFF
--- a/mitxgraders/helpers/calc/expressions.py
+++ b/mitxgraders/helpers/calc/expressions.py
@@ -414,7 +414,7 @@ class MathParser(object):
         #   subscripts (optional):
         #       any combination of alphanumeric and underscores
         #   lower_indices (optional):
-        #       Of form "_{(-)<alaphnumeric>}"
+        #       Of form "_{(-)<alphanumeric>}"
         #   upper_indices (optional):
         #       Of form "^{(-)<alphanumeric>}"
         #   tail (optional):
@@ -608,7 +608,7 @@ class MathExpression(object):
         bad_suffixes = set(suff for suff in self.suffixes_used if suff not in suffixes)
         if bad_suffixes:
             bad_suff_names = ', '.join(sorted(bad_suffixes))
-            message = "Invalid Input: {} not permitted directly after a number."
+            message = "Invalid Input: {} not permitted directly after a number"
 
             # Check to see if there is a corresponding variable name
             if any(suff in variables for suff in bad_suffixes):
@@ -627,6 +627,21 @@ class MathExpression(object):
             raise UndefinedFunction(message.format(bad_suff_names))
 
     def eval(self, variables, functions, suffixes, allow_inf=False):
+        """
+        Numerically evaluate a MathExpression's tree, returning a tuple of the
+        numeric result and evaluation metadata.
+
+        Also recasts some errors as CalcExceptions (which are student-facing).
+
+        Arguments:
+            variables (dict): maps variable names to values
+            functions (dict): maps function names to values
+            suffixes (dict): maps suffix names to values
+            allow_inf (bool): If true, any node evaluating to inf will throw
+                a CalcOverflowError
+
+        See class-level docstring for example usage.
+        """
         self.check_scope(variables, functions, suffixes)
 
         # metadata_dict['max_array_dim_used'] is updated by eval_array
@@ -670,8 +685,8 @@ class MathExpression(object):
     @staticmethod
     def eval_node(node, actions, allow_inf):
         """
-        Recursively evaluates a node, calling itself on the nodes children. Delegates to one of the provided actions,
-        passing evaluated child nodes as arguments.
+        Recursively evaluates a node, calling itself on the node's children.
+        Delegates to one of the provided actions, passing evaluated child nodes as arguments.
         """
 
         if not isinstance(node, ParseResults):
@@ -1096,6 +1111,7 @@ class MathExpression(object):
         return result
 
 PARSER = MathParser()
+
 def evaluator(formula,
               variables=DEFAULT_VARIABLES,
               functions=DEFAULT_FUNCTIONS,
--- a/mitxgraders/helpers/calc/expressions.py
+++ b/mitxgraders/helpers/calc/expressions.py
@@ -1020,10 +1020,10 @@ class MathExpression(object):
         >>> MathExpression.eval_parallel([1,1])
         0.5
         >>> MathExpression.eval_parallel([1,0])
-        nan
+        0
         """
         if 0 in parse_result:
-            return float('nan') # Thinking about circuits, should x || 0 = 0 ?
+            return 0
         reciprocals = [1. / num for num in parse_result]
         return 1. / sum(reciprocals)
 
--- a/mitxgraders/helpers/calc/expressions.py
+++ b/mitxgraders/helpers/calc/expressions.py
@@ -392,7 +392,6 @@ class MathParser(object):
             Combine(
                 inner_number +
                 Optional(CaselessLiteral("E") + Optional(plus_minus) + number_part),
-                adjacent=False
             )("num")
             + Optional(suffix)("suffix")
         )("number")
--- a/mitxgraders/helpers/calc/expressions.py
+++ b/mitxgraders/helpers/calc/expressions.py
@@ -3,7 +3,7 @@ expressions.py
 
 Defines classes to parse and evaluate mathematical expressions. Implements
 similar functionality and API as edX's calc.py, but re-written with enhancements
-and to betterseparate parsing and evaluation.
+and to better separate parsing and evaluation.
 
 To evaluate a mathematical expression like '2^a + [x, 2]*[y, 4]', we proceed
 in two steps:
@@ -1005,7 +1005,7 @@ class MathExpression(object):
         """
         Operator associated with parallel resistors (it's commutative).
 
-        Return NaN if there is a zero among the inputs.
+        Return 0 if there is a zero among the inputs.
 
         Arguments:
             parse_result: A list of numbers to combine appropriately
--- a/mitxgraders/helpers/calc/expressions.py
+++ b/mitxgraders/helpers/calc/expressions.py
@@ -1,11 +1,37 @@
 """
 calc.py
 
-Parser and evaluator for mathematical expressions.
+Defines classes to parse and evaluate mathematical expressions.
 
-Uses pyparsing to parse. Main function is evaluator().
+To evaluate a mathematical expression like '2^a + [x, 2]*[y, 4]', we proceed
+in two steps:
 
-Heavily modified from the edX calc.py
+    1. First, the expression is parsed into a tree
+
+                               /--- NUMBER --- '2'
+                 /--- POWER ---
+                /              \--- VARIABLE --- 'a'
+               /
+              /
+        SUM --  -- OP -- '+'
+              \                                   /--- VARIABLE --- 'x'
+               \                    /--- ARRAY ---
+                \                  /              \--- NUMBER --- '2'
+                 \                /
+                  \--- PRODUCT ---  -- OP -- '*'
+                                  \
+                                   \              /--- VARIABLE --- 'y'
+                                    \--- ARRAY ---
+                                                  \--- NUMBER --- '4'
+
+    2. Next, the tree is evaluated from the leaves upwards.
+
+This file defines two main classes:
+ - MathParser, used to parse mathematical strings into a tree
+ - MathExpression, holds the parse tree for a given mathematical expression
+   and can be used to evaluate the tree with a given scope.
+and a function:
+ - evaluator: a convenience function that parses and evaluates strings.
 """
 from __future__ import division
 import copy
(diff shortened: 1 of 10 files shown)


ChristopherChudzicki:
@jolyonb This is ready for review.

I had to specify pytest-cov version in requirements.txt. Newest version (2.6) dropped support for Python <3.4. We have the most recent version that supports Python 2.7.

https://pytest-cov.readthedocs.io/en/latest/changelog.html#id1

tests/helpers/calc/test_expressions.py
```diff
@@ -1,5 +1,5 @@
 """
-Tests of calc.py
+Tests of expression.py that arent covered elsewhere
```
jolyonb: Nit: `aren't`

tests/formulagrader/test_formulagrader.py
```diff
@@ -94,10 +94,14 @@ def test_fg_invalid_input():
     with raises(CalcError, match=expect):
         grader(None, "R")
 
-    expect = "Invalid Input: Could not parse '5pp' as a formula"
+    expect = "Invalid Input: pp not permitted directly after a number."
     with raises(CalcError, match=expect):
         grader(None, "5pp")
 
+    expect = "Invalid Input: m not permitted directly after a number. \(did you forget to use \* for multiplication?"
```
jolyonb: Nit: Does the parenthesis close? Should d in "did" be capitalized?

tests/helpers/calc/test_expressions.py
```diff
@@ -189,3 +194,11 @@ def test_div_by_zero():
         evaluator("1/0")
     with raises(CalcZeroDivisionError, match=msg):
         evaluator("0^-1")
+
+def test_suffix_capitalization_error():
+    variables = {}
+    functions = {}
+    suffixes = {'M'}
+    match = "Invalid Input: m not permitted directly after a number. \(did you mean M\?\)"
```
jolyonb: Nit: "D" in did again.

mitxgraders/helpers/calc/expressions.py
```diff
@@ -1,11 +1,37 @@
 """
 calc.py
```
jolyonb: No longer calc.py!

mitxgraders/helpers/calc/expressions.py
```diff
@@ -1062,3 +1085,93 @@ def eval_sum(self, parse_result):
             else:
                 raise CalcError("Unexpected symbol {} in eval_sum".format(op))
         return result
+
+PARSER = MathParser()
```
jolyonb: This seems somewhat hidden here, given that it's the fundamental object in this module. Perhaps it's worth mentioning in the docstring?

mitxgraders/helpers/calc/expressions.py
jolyonb: May also want to mention that this implements similar functionality to calc.py in edX, but has been completely rewritten with new structures.

mitxgraders/helpers/calc/expressions.py
```diff
@@ -381,57 +267,69 @@ def cast_np_numeric_as_builtin(obj, map_across_lists=False):
...
+    be evaluated.
+
+    Usage
+    =====
+    >>> new_parser = MathParser()
+    >>> parsed = new_parser.parse('2*x + 5')
+    >>> isinstance(parsed, MathExpression)
+    True
+    >>> parsed
+    <BLANKLINE>
```
jolyonb: This is curious? What's with the <BLANKLINE>?

mitxgraders/helpers/calc/expressions.py
```diff
@@ -381,57 +267,69 @@ def cast_np_numeric_as_builtin(obj, map_across_lists=False):
...
         """
-        When a variable is recognized, store it in `variables_used`.
+        When pyparsing encounters a variable, store it in variables_used
         """
         self.variables_used.add(tokens[0][0])
 
     def function_parse_action(self, tokens):
         """
-        When a function is recognized, store it in `functions_used`.
+        When pyparsing encounters a function, store it in variables_used
```
jolyonb: Store it in functions instead of variables?

mitxgraders/helpers/calc/expressions.py
```diff
@@ -381,57 +267,69 @@ def cast_np_numeric_as_builtin(obj, map_across_lists=False):
...
         """
         self.functions_used.add(tokens[0][0])
 
     def suffix_parse_action(self, tokens):
         """
-        When a suffix is recognized, store it in `suffixes_used`.
+        When pyparsing encounters a suffix, store it in variables_used
```
jolyonb: Similarly, copy-paste error on variables_used.

mitxgraders/helpers/calc/expressions.py
```diff
@@ -490,7 +388,7 @@ def parse_algebra(self):
             + Optional(suffix)("suffix")
```
jolyonb: I highly approve of taking out the suffix matching at this stage, and just matching anything. I'm guessing that this will match something like `1.5e-3%`, which is now ambiguous? Also, will this match things like `1.5m%m%mmp`? Probably the price to pay in order to give a good error message...

mitxgraders/helpers/calc/expressions.py
```diff
@@ -469,16 +367,16 @@ def parse_algebra(self):
...
-        # Construct a number as a group consisting of a text string (num) and an optional
-        # suffix num can include a decimal number and numerical exponent, and can be
+        # Construct number as a group consisting of a text string ("num") and an optional suffix.
+        # num can include a decimal number and numerical exponent, and can be
         # converted to a number using float()
-        # suffix is the suffix string that matches one of our suffixes
+        # suffix may contain alphas or %
         # Spaces are ignored inside numbers
         # Group wraps everything up into its own ParseResults object when parsing
         number = Group(
```
jolyonb: Do we have a good reason to have `adjacent=False` here? I suspect I have a strong preference for True...

mitxgraders/helpers/calc/expressions.py
```diff
@@ -381,57 +267,69 @@ def cast_np_numeric_as_builtin(obj, map_across_lists=False):
...
-        """
-        Create the ParseAugmenter for a given math expression string.
 
-        Do the parsing later, when called like `OBJ.parse_algebra()`.
-        """
-        self.math_expr = math_expr
-        self.tree = None
+    def __init__(self):
+        self.cache = {}
+        self.grammar = self.get_grammar()
```
jolyonb: I approve of constructing and storing this!

mitxgraders/helpers/calc/expressions.py
```diff
@@ -509,52 +407,51 @@ def parse_algebra(self):
         #   lower_indices (optional):
         #       Of form "_{(-)<alaphnumeric>}"
         #   upper_indices (optional):
-        #       Of form "^{(-)<alaphnumeric>}"
+        #       Of form "^{(-)<alphanumeric>}"
```
jolyonb: Good catch!

mitxgraders/helpers/calc/expressions.py
```diff
@@ -509,52 +407,51 @@ def parse_algebra(self):
         #   lower_indices (optional):
         #       Of form "_{(-)<alaphnumeric>}"
```
jolyonb: Can you get this one too?

mitxgraders/helpers/calc/expressions.py
```diff
@@ -575,90 +472,101 @@ def parse_algebra(self):
...
-            feed it the output of `handle_node` for each child node.
-            """
-            if not isinstance(node, ParseResults):
-                # Entry is either a (python) number or a string.
-                # Return it directly to the next level up.
-                return cast_np_numeric_as_builtin(node)
+        try:
+            BracketValidator.validate(expression)
+            tree = self.grammar.parseString(expression)[0]
+            parsed = MathExpression(expression,
```
jolyonb: I really like this. Having separated out the suffixes, tree parsing is indeed general.

mitxgraders/helpers/calc/expressions.py
```diff
@@ -575,90 +472,101 @@ def parse_algebra(self):
...
+class MathExpression(object):
+    """
+    Holds the parse tree for mathematical expression; returned by MathParser.
+
+    Attributes:
+        - expression (str): the original string that generated this parse tree
+        - variables_used (set)
+        - functions_used (set)
+        - suffixes_used (set)
+
```
jolyonb: Is there a reason not to have `max_array_dim_used` here too? Maybe because we don't get that from just parsing; we compute it from an evaluation...

tests/helpers/calc/test_expressions.py
jolyonb: Actually, keeping inline with other things, maybe just drop the period after "number".

tests/formulagrader/test_formulagrader.py
jolyonb: As with other comment - drop the period after "number" instead.

mitxgraders/helpers/calc/expressions.py
```diff
@@ -689,11 +596,109 @@ def check_variables(self):
 
             raise UndefinedFunction(message.format(funcnames))
 
+        bad_suffixes = set(suff for suff in self.suffixes_used if suff not in suffixes)
+        if bad_suffixes:
+            bad_suff_names = ', '.join(sorted(bad_suffixes))
+            message = "Invalid Input: {} not permitted directly after a number."
```
jolyonb: Here we go. Remove the period after "number"

mitxgraders/helpers/calc/expressions.py
```diff
@@ -689,11 +596,109 @@ def check_variables(self):
 
             raise UndefinedFunction(message.format(funcnames))
 
+        bad_suffixes = set(suff for suff in self.suffixes_used if suff not in suffixes)
```
jolyonb: :+1: for this treatment!

mitxgraders/helpers/calc/expressions.py
ChristopherChudzicki: It's a blank line :).

I have no idea what's causing that. It happens in master branch, too. Atom and iTerm2 print it as an actual blank line, except when run in a doctest. Doctest shows it as a <BLANKLINE>. 

Anyway, it never reared its head before because we didn't have a test for FormulaParser.dump_parse_result (it had `pragma: no-cover`). But I wrote a test for the printing, so now we see it.

But I don't know where it comes from ... Maybe built in to pyparsing's `asxML()` method?

mitxgraders/helpers/calc/expressions.py
```diff
@@ -689,11 +596,109 @@ def check_variables(self):
...
+                for suff1 in suffixes:
+                    if suff2.lower() == suff1.lower():
+                        caselist.add(suff1)
+            if len(caselist) > 0:
+                betternames = ', '.join(sorted(caselist))
+                message += " (did you mean " + betternames + "?)"
+
+            raise UndefinedFunction(message.format(bad_suff_names))
+
+    def eval(self, variables, functions, suffixes, allow_inf=False):
```
jolyonb: This one could really use a docstring.

mitxgraders/helpers/calc/expressions.py
```diff
@@ -689,11 +596,109 @@ def check_variables(self):
...
+
     # The following functions define evaluation actions, which are run on lists
     # of results from each parse component. They convert the strings and (previously
     # calculated) numbers into the number that component represents.
 
-    def eval_number(self, parse_result):
+    @staticmethod
+    def eval_node(node, actions, allow_inf):
+        """
+        Recursively evaluates a node, calling itself on the nodes children. Delegates to one of the provided actions,
```
jolyonb: Nit: `node's`

mitxgraders/helpers/calc/expressions.py
ChristopherChudzicki: Re `1.5e-3%` Yes, that matches. And if `%` is defined as `0.01`, it evaluates to `1.5e-05` as it should.

Re `1.5m%m%mmp` Yes, that matches, but you do at least get a (somewhat) helpful error message: `'Invalid Input: m%m%mmp not permitted directly after a number`'

**But**: No useful error message is given for `5x_1`, since suffixes match `Word(alphas + '%')`. I briefly tried to make suffixes match all variables, but abandoned that due to time. Can come back to it later if we want, prefer to merge this without it.

mitxgraders/helpers/calc/expressions.py
```diff
@@ -965,38 +989,37 @@ def eval_parallel(self, parse_result):
...
-        >>> parser.eval_parallel([1,1])
+        >>> MathExpression.eval_parallel([1,1])
         0.5
-        >>> parser.eval_parallel([1,0])
+        >>> MathExpression.eval_parallel([1,0])
         nan
         """
         if 0 in parse_result:
-            return float('nan')
+            return float('nan') # Thinking about circuits, should x || 0 = 0 ?
```
jolyonb: This seems reasonable to me, actually. Let's do this.

mitxgraders/helpers/calc/expressions.py
```diff
@@ -965,38 +989,37 @@ def eval_parallel(self, parse_result):
...
+    def eval_product(parse_result):
         """
         Multiply/divide inputs appropriately
 
         Arguments:
             parse_result: A list of numbers to combine, separated by "*" and "/"
             [a, "*", b, "/", c] = a*b/c
 
-        Has some extra logic to avoid ambiguous vector tirple products.
+        Has some extra logic to avoid ambiguous vector triple products.
```
jolyonb: Nice :-)

mitxgraders/helpers/calc/expressions.py
jolyonb: Let's call it a bugfix. I'd like to train people not to write `1.5 e - 2` 😄

mitxgraders/helpers/calc/expressions.py
ChristopherChudzicki: Yes, that is why.

IMO, `max_array_dim_used` is weird as heck since it is based on evaluation but triggered by parsing into an ArrayNode. I mean, I do think that it behaves the way we want, I just think it's weird.

Consider:

```python
from mitxgraders.helpers.calc import evaluator, MathArray

A = MathArray([1, 2, 3])

result_1, meta_1 = evaluator('A', {'A': A})
print(meta_1.max_array_dim_used) # 0

result_1, meta_1 = evaluator('[A]', {'A': A})
print(meta_1.max_array_dim_used) # 2

```

jolyonb: Ok, that example gave me a "Wait, what?" for a few seconds. I believe this is actually exactly the behavior that we want, though it is somewhat counter-intuitive. If you wanted `[A]` to have dimension one, that would actually be much harder for us to implement.

jolyonb:
Feel free to mark any unconstructive comments as resolved, btw.

ChristopherChudzicki:
I believe I've just resolved everything except for the `adjacent` issue. Just confirm you want that and I can change it, or merge as is.

mitxgraders/helpers/calc/expressions.py
```diff
@@ -1,7 +1,9 @@
 """
-calc.py
+expressions.py
 
-Defines classes to parse and evaluate mathematical expressions.
+Defines classes to parse and evaluate mathematical expressions. Implements
+similar functionality and API as edX's calc.py, but re-written with enhancements
+and to betterseparate parsing and evaluation.
```
jolyonb: space between better and separate

mitxgraders/helpers/calc/expressions.py
```diff
@@ -996,10 +1020,10 @@ def eval_parallel(parse_result):
         >>> MathExpression.eval_parallel([1,1])
         0.5
         >>> MathExpression.eval_parallel([1,0])
-        nan
+        0
```
jolyonb: Now need to update the docstring to match this behavior.

jolyonb: Line 1008, I think?

mitxgraders/helpers/calc/expressions.py
ChristopherChudzicki: you didn't want to tell me to better separate those words?

jolyonb:
If we don't break any tests by setting `adjacent` to True, then let's go ahead and do it.

mitxgraders/helpers/calc/expressions.py
ChristopherChudzicki: done :)

mitxgraders/helpers/calc/expressions.py
ChristopherChudzicki: done

jolyonb:
Awesome job. This almost feels like it's worthy of a revision number!

Issue #131, opened 2018-08-28
Negative Powers to Zero-Det Matrix Not Handled


jolyonb:
If you raise a matrix with zero determinant to a negative power, the raw error from `np.linalg.pow` is raised, rather than catching the error beforehand and presenting a nice error message.

ChristopherChudzicki:
Good point. Probably a simple try/catch in MathArray's pow method.

ChristopherChudzicki:
@jolyonb We had talked about rolling this into #134, but I did not address it yet. Partly because it seems separate issue, but mostly because #134 is ready for review and I should work on my classes :)

But if this is causing problems for anyone you know of, let me know and I'll make the time soon.

jolyonb:
No, no rush at all.

ChristopherChudzicki:
@jolyonb I should have deleted that comment :) I decided it was easy and so I did it. #136

jolyonb:
Ah, I started with the earliest email from you and began working my way up ;-)

PR #136, opened 2018-09-16
implement try/catch for singular matrices in __pow__

ChristopherChudzicki:
Addresses #131 

@jolyon Nearly singular things might still behave badly? You probably know better than me...


jolyonb:
Nearly-singular can still behave badly, but for the purposes of numerics, they're essentially singular. This should cover all bases.

Issue #138, opened 2018-09-20
Add braces to bracket validator


jolyonb:
It occurred to me that variable names can now sometimes have braces `name_{stuff}`, so we should probably be validating braces as will in our bracket validator.

jolyonb:
Resolved by #139

Issue #133, opened 2018-09-16
atan2


jolyonb:
Would be good to have an arctangent function that accepts an x and a y value, capable of returning results in all four quadrants.

ChristopherChudzicki:
Sure!  Is this just “it would be nice” or is there a class you have a specific purpose in mind for?  Do you care what the name is? We could overload ‘arctan(...)’ so that except one or two arguments (This is what Mathematica does) or have a separate function.

jolyonb:
I'm not sure we have the infrastructure to overload a function at this stage? I'm not sure we really need it, either... `atan2` is a fine function name (used by various codes). And yes, I have an explicit use for it in computing quantum phases.

jolyonb:
Resolved by #139

PR #139, opened 2018-09-21
add arctan2 and curly brace validation

ChristopherChudzicki:
Edit: This PR addresses #133 and #138


jolyonb:
Want to update version numbers everywhere to 1.2.1? I think we should also include #138 in a version update.

Otherwise, this looks fine to me. Might suggest adding a couple of simple tests in the functions file where we test the inverse trig functions - arctan2(x, y) == arctan(y/x) type things.

ChristopherChudzicki:
Sure, I can implement #138 in this PR and add it to changelog. Either later tonight or tomorrow.

I just updated version in README.

I moved the arctan2 tests into their own file because of https://bugs.python.org/issue6835

jolyonb:
Looks good. Also remember to update `__version__` in `mitxgraders/version.py`

mitxgraders/helpers/calc/expressions.py
```diff
@@ -203,24 +207,28 @@ def raise_open_without_close(formula, stack):
...
+            if num_opened:
+                still_open[key] = num_opened
+
+        # sort so error messages come in definite order, for testing purposes
+        sorted_still_open = sorted(still_open.keys(), key=lambda x: bracket_registry[x].name)
+
+        if still_open:
+            message = "Invalid Input:\n"
+            for key in sorted_still_open:
+                message += ('{count} {name} were opened without being closed (highlighted below)\n'
```
ChristopherChudzicki: I mangled the english a bit here (plural/singular not accounted for), but I don't think it matters and this is easier to maintain/extend.

jolyonb: Was noticing this. You could add a "plural_name" to the dictionary, and put an if statement here on `count`.

jolyonb: Given that it's a handful of lines and that this is unlikely to ever be touched again, might as well go for it, I think.

ChristopherChudzicki: 👍

jolyonb:
All looks good 👍

PR #140, opened 2018-09-21
Creating a tidy script

jolyonb:
Two parts to this one.

First, a python script that uploads a tar.gz file to edX. You need to create a `config.py` file with your credentials to do this, following the example file. I made this for another course, and figured it would be useful here.

Second, a shell script that just does all of our general updates in one go (making the github docs, making the zip file, and updating the course with the latest examples and zip file).


jolyonb:
Apparently the library that I'm using to do the upload has a tonne of requirements though... lol.

ChristopherChudzicki:
Good idea! One quesstion: `retquirements.txt` expanded a bunch, was all of that necessary? E.g., what is `six`?

jolyonb:
Alas, yes. The `requests` library that I loaded to do the http stuff depends on a bunch of stuff - all of those extra things were in the dependency tree of `requests`. I believe that `six` is a compatibility library for writing code that works in python 2 and python 3.

PR #145, opened 2018-10-06
Wrap all variables in invisibrackets to stop breaking

### LINKED ISSUE #144: Asciimath and apostrophes in variable names
The following expression looks terrible in asciimath: `x'/x`. Similarly bad is `x''/x`. Can you think of a way to fix this in the preprocessor?

jolyonb:
Resolves #144.


jolyonb:
Your test should fail, because the arguments won't be wrapped.

jolyonb:
Which actually brings up a different point: should functions be wrapped, so that something like `p(x)/2` doesn't look terrible?

mitxgraders-js/MJxPrep.test.js
```diff
@@ -105,6 +105,12 @@ describe('preProcessEqn', () => {
     const eqn = 'cross(x) + cross(a + b, c) + y'
     expect(preProcessEqn(eqn)).toBe('cross(x) + {:(a + b) times c:} + y')
   } )
+
+  it('wraps primed expressions in unbreakable invisible brackets', () => {
+    const expr = "f''(x)/f(x) + g_a'(y) "
+    expect(preProcessEqn(expr)).toBe("{:f''(x):}/f(x) + {:g_a'(y):}")
```
ChristopherChudzicki: I just added this test case. Based on the behavior of [http://asciimath.org/](http://asciimath.org/), I am pretty sure this test does describe the desired behavior.

Currently this test fails:
```
Expected: "{:f''(x):}/f(x) + {:g_a'(y):}"
Received: "{:f'':}(x)/f(x) + g_{:a':}(y) "
```
(I just noticed the extra space at the end in definition of `expr`, but that's not the main issue.)

Do you want to try and fix it? I can also work on it, but probably not till Wednesday or Thursday.

ChristopherChudzicki: Oh, I see. You're saying this test fails because of a different issue?

Re:
> Which actually brings up a different point: should functions be wrapped, so that something like p(x)/2 doesn't look terrible?

Yes, that sounds reasonable. And it might make a whole bunch of our existing preprocessing redundant?

Basically:

- anything of the for `[non-operator characters](...)` must be a function call
- and so should be wrapped

right?

ChristopherChudzicki:
Really, shouldn't our rule for wrapping just be:

- anything that parses to a valid variable name is wrapped
- anything that parses to a function call is wrapped

This might result in some unnecessary wrapping, but is that bad?

mitxgraders-js/MJxPrep.test.js
jolyonb: Yeah, I think that's the case. Would probably be good to have all functions be wrapped (I don't see how this could break anything?), which would look after the prime problem for functions. However, there's still a prime problem for variables. We can modify the regex to use lookahead to see if the variable is proceeded by a (, in which case it's a function and can be ignored.

Why don't I modify this to fix the variable prime problem only, and leave you to fix the function problem in a separate PR?

jolyonb:
Yeah, that would be best. That will require a bunch of extra parsing to handle numbered variables correctly (I'm looking at you, minus sign!), unfortunately...

I can't think of why extra wrapping would be bad.

Ok, this just got harder than I was hoping for. Can I leave you to look after it? I don't need this to be working until Wednesday 10/17, when I'm running a workshop at Harvard.

mitxgraders-js/MJxPrep.test.js
ChristopherChudzicki: > Why don't I modify this to fix the variable prime problem only, and leave you to fix the function problem in a separate PR?

That sounds good. I'm not sure if you need to modify anything other than my broken test :)

jolyonb: I need to detect that I'm not in a function ;-)

ChristopherChudzicki:
> Ok, this just got harder than I was hoping for. Can I leave you to look after it? I don't need this to be working until Wednesday 10/17, when I'm running a workshop at Harvard.

That sounds good. I will take a much closer look at this on Wednesday. If I can't get the ideal fix ready by your workshop, we can at least do the original fix.

jolyonb:
Ok, I've updated it so that anything that parses as a valid variable name gets wrapped. This was a pain to figure out, alas. The regex is a mess, because javascript's regex doesn't allow for atomic groupings (useful because they stop backtracking, which means that something like `f0(x)` with a negative lookahead on `(` doesn't match `f`). I added a whole bunch of tests. Please look closely, as this mad a mess of some other tests too.

We have to decide if we want variables and functions to be wrapped at the top or the bottom of the processor. I put it at the bottom, but now I'm leaning towards having it at the top, so that things like ^T and dagger don't get wrapped, which makes a mess of the tests... Let me know what you think.

ChristopherChudzicki:
This seems good to me. I agree the regexp is a mess, but your comments help a lot!

Re:

> We have to decide if we want variables and functions to be wrapped at the top or the bottom of the processor. I put it at the bottom, but now I'm leaning towards having it at the top, so that things like ^T and dagger don't get wrapped, which makes a mess of the tests... Let me know what you think.

Let's just keep it where it is for now. Moving the variable wrapping to top could be a little more convenient, but I do think the (not-yet-implemented) function call wrapping should be at the bottom to avoid wrapping function-calls that will be specially transformed (e.g., `cross(a, b)`, `trans(A)`, etc.

PR #146, opened 2018-10-10
wrap function calls in unbreakable invisible braces

ChristopherChudzicki:
Does what it says, and reorganized javascript tests slightly.

I updated the edge course to see it live... perhaps I should have waited. The uploader script is really great.

### DIFF
--- a/mitxgraders-js/MJxPrep.test.js
+++ b/mitxgraders-js/MJxPrep.test.js
@@ -6,6 +6,8 @@ const {
   groupExpr,
   shallowListSplit,
   preProcessEqn,
+  wrapVariables,
+  wrapFuncCalls,
   columnizeVectors,
   funcToPostfix
 } = window.MJxPrepExports
@@ -65,7 +67,7 @@ describe('preProcessEqn', () => {
   it('replaces log10 and log2', () => {
     const expr = 'log10(1 + log2(x))'
     const result = preProcessEqn(expr)
-    expect(result).toBe('log_10(1+log_2({:x:}))')
+    expect(result).toBe('{:log_10(1+{:log_2({:x:}):}):}')
   } )
 
   it('replaces fact and factorial', () => {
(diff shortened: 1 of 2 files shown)


mitxgraders-js/MJxPrep.test.js
```diff
@@ -103,12 +105,24 @@ describe('preProcessEqn', () => {
...
+  } )
+
+} )
+
+describe('wrapVariables', () => {
+
   it('wraps variable names in unbreakable invisible brackets', () => {
     const exprs = [
-      ["f''/f + g_a'+sin(horse)", "{:f'':}/{:f:}+{:g_a':}+sin({:horse:})"],
+      ["f''/f + g_a' + sin(horse)", "{:f'':}/{:f:} + {:g_a':} + sin({:horse:})"],
```
jolyonb: Wait, how does this work? I thought I stripped spaces out of the entry so that it wouldn't mess with the regex?

ChristopherChudzicki: Note that I moved your variable wrapping into `wrapVariables`, and this only tests the `wrapVariables` function. So `preprocessEqn` still strips whitespace before wrapping the variables.

That said, what part of `preprocessEqn` requires whistepsace-stripping? Do we really need to strip whitespace? I am not sure...

jolyonb: How do you think `f (x)` should render? What about `f_{mu nu}?` There are just a bunch of places where a space might be inserted, and I didn't want to deal with this in the regex. Besides, the first thing we do in the processing is to strip spaces... Removing them just made my thought processes much more concrete.

jolyonb:
Merging. Feel free to respond to above comment though. Please run tidy again!

mitxgraders-js/MJxPrep.test.js
ChristopherChudzicki: I think you're right. Stripping whitespace will make `f (x)/2` render correctly, so let's stick with it.

Issue #150, opened 2018-10-17
Read answer from "expect" field


jolyonb:
If there is only a single textbox entry, it would be nice to be able to read the answer from the expect field, so it doesn't need to be included in the edX code twice.

jolyonb:
Resolved by #151

PR #151, opened 2018-10-17
Infer expect

ChristopherChudzicki:
This addresses #150 for ItemGraders only. Implementing this for ItemGrader only was very easy because ItemGraders have a strictly specified answers schema.

I suspect we could do something that works in general, but I would need to think a bit more about it.

I wrote this fairly quickly in case you wanted it for today's workshop. I think it should be safe to merge, but feel free to reject if you want a more general solution =).

### DIFF
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -92,26 +92,28 @@ class AbstractGrader(ObjectWithSchema):
         Used by edX as the check function (cfn).
 
         Arguments:
-            expect: The value of edX customresponse expect attribute (ignored)
+            expect: The value of edX customresponse expect attribute (ignored).
             student_input: The student's input passed by edX
 
         Notes:
-            This function ignores the value of expect.
+            This function ignores the value of expect. The expect argument is
+            provided because edX requires that a check function to have the
+            signature above.
 
-            The expect argument is provided because edX expects a check function
-            to have the signature above. Our graders usually ignore the first
-            argument, expect. Instead, we usually pass None as the expect value,
-            indicating that the grader should instead read the expected answer
-            from the grader's configuration. This is because we generally use
+            Our graders usually read the author's expected answer from the
+            grader configuration. This is because we generally use
             dictionaries to store the expected input along with correctness,
             grades, and feedback messages.
 
-            Note that authors should still specify the <customresponse />
+            Authors should still specify the <customresponse />
             expect attribute because its value is displayed to students as
             the "correct" answer.
 
-            Note: ItemGraders can infer its answers from the their answers from
-            the expect value of a customresponse tag.
+            ItemGraders: If no answer is provided in the configuration, an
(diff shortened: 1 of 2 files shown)


mitxgraders/baseclasses.py
```diff
@@ -98,19 +98,20 @@ def __call__(self, expect, student_input):
         Notes:
             This function ignores the value of expect.
```
jolyonb: I think some more rewording is necessary here...

mitxgraders/baseclasses.py
```diff
@@ -98,19 +98,20 @@ def __call__(self, expect, student_input):
...
+            dictionaries to store the expected input along with correctness,
+            grades, and feedback messages.
+
+            Note that authors should still specify the <customresponse />
             expect attribute because its value is displayed to students as
             the "correct" answer.
 
-            The answer that we pass to check is None, indicating that the
-            grader should read the answer from its internal configuration.
+            Note: ItemGraders can infer its answers from the their answers from
```
jolyonb: Change to "If no answer is provided in the configuration, an ItemGrader will attempt to infer its answer from the expect parameter of a textline or CustomResponse tag. Note that this doesn't work when an ItemGrader is embedded inside a ListGrader."

jolyonb:
This is exactly what I was after -- I'm a bit surprised it was this simple though! I found that I've been copying code a lot and just changing the "answers" tag, which was getting really annoying. Also, copy paste errors where I put the answer in the expect but forgot to change the answers tag... We should update documentation to reflect this too, but I'm happy with this for the moment.

mitxgraders/baseclasses.py
ChristopherChudzicki: I just reworded the comments a bit. I left the part about `expect` being ignored: It is true that `AbstractGrader.__call__` completely ignores `expect`. I agree this is a little weird... feel free to word something better :)

ChristopherChudzicki:
> I found that I've been copying code a lot and just changing the "answers" tag, which was getting really annoying

Yeah, I can see that. For all the work I've done on this library, I've probably only written a dozen problems outside the demo course, so I forget about workflow issues sometimes 😦 

> We should update documentation to reflect this too

Yes!

mitxgraders/baseclasses.py
```diff
@@ -92,25 +92,28 @@ def __call__(self, expect, student_input):
...
             expect attribute because its value is displayed to students as
             the "correct" answer.
 
-            The answer that we pass to check is None, indicating that the
-            grader should read the answer from its internal configuration.
+            ItemGraders: If no answer is provided in the configuration, an
+            ItemGrader will attempt to infer its answer from the expect
+            parameter of a textline or CustomResponse tag. Note that this does
+            not work when an ItemGrader is embedded inside a ListGrader. See
+            ItemGrader.__call__ the implementation.
```
jolyonb:
```suggestion
            ItemGrader.__call__ for the implementation.
```

PR #152, opened 2018-10-18
Fixing Unicode

jolyonb:
So, it turns out that edX sends us a unicode version of the expect string, not a string version. This fixes the appropriate pieces and makes everything work on edX :-)

### DIFF
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -425,6 +425,6 @@ class ItemGrader(AbstractGrader):
         grader configuration.
         """
         if not self.config['answers'] and expect is not None:
-            self.config['answers'] = self.schema_answers(expect)
+            self.config['answers'] = self.schema_answers(expect.encode('utf-8'))
 
         return super(ItemGrader, self).__call__(expect, student_input)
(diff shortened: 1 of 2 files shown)


jolyonb:
With this fix, the following code works in edX!!!

```xml
<problem>
<script type="loncapa/python">
from mitxgraders import *
</script>

<customresponse cfn="FormulaGrader(variables=['x'])" answer="2*x" >
  <textline math="true" />
</customresponse>

</problem>```

Putting primes in variable names is a bit annoying, alas - you need to do XML escaping. However, for simple things, this is beautiful!

**Comment by jolyonb** [comment]
Time to do some large scale restructuring of the documentation, I guess...

**Comment by ChristopherChudzicki** [comment]
This is cool! 

Re:
> Time to do some large scale restructuring of the documentation, I guess...

It probably boils down to personal preference... I think that if I were writing a large number of problems, I would try to use essentially the same customresponse tag for all problems, and have the tag attributes reference python variables `$grader` and `$expect`. Something like:

```xml
<problem>
<script type="loncapa/python">
from mitxgraders import *
expect = 'x^2'
grader = FormulaGrader(answers=expect, variables=['x'])
</script>

<customresponse cfn="$grader" expect="$expect">
  <textline math="true" />
</customresponse>

</problem>
```
If you decided to do this, we could even expose `grader.config['answers'][0]['expect']` as `grader.expect` (if it exists and is unique, or expose the single answer with `'ok': True`).

PR #153, opened 2018-10-30
Fixing error with single input in integral grader

jolyonb:
If only a single input is given by a student in an integralgrader, we need to wrap that input in a list.


mitxgraders/plugins/integralgrader.py
```diff
@@ -351,6 +351,10 @@ def structure_and_validate_input(self, student_input):
     def check(self, answers, student_input, **kwargs):
         """Validates and cleans student_input, then checks response and handles errors"""
         answers = self.config['answers'] if answers is None else answers
+        # If only a single input has been provided, wrap it in a list
+        # This is possible if only the integrand is required from the student
```
ChristopherChudzicki: This looks good, I just want to make sure I understand:
 - For `integral(f(x), x, a, b)`, if only integrand is required, author could just use `FormulaGrader`. 
 - But author might want to use `IntegralGrader` so that other equivalent integrands are accepted, like `f(x) + g(x)`, where `integral(g(x), x, a, b)=0`.
Is that right?

ChristopherChudzicki: Also, should we add a test for this in `test_integralgrader`? I don't think reports coverage decrease as failures anymore (I forget why we changed that).

I can do that after class today, or you can, either way.

jolyonb: The situation I needed this was the integral `int(e^(-u^2), u, 1, infty)`. You can make the substitution `x=u^2` without changing the limits, so I wanted both to be accepted.

I'm not sure why coverage decreases aren't reporting as failures. I didn't change that...

And yes, this should be tested. But you beat me to it ;-)

PR #154, opened 2018-11-06
Updating documentation

jolyonb:
This updates the documentation to discuss the possibility of not providing an `answers` key. It also expands on the discussion of edX syntax.


docs/grading_lists/list_grader.md
```diff
@@ -14,7 +14,7 @@ grader = ListGrader(
 )
 ```
 
-Each element of answers is set as an answer that is passed as the answers key into the subgrader. This should be set up as two input boxes that the student types in.
+Each element of answers is set as an answer that is passed as the answers key into the subgrader. This should be set up as two input boxes that the student types in. Note that the `answers` key is provided as a python list for a `ListGrader`, and is a required key.
```
ChristopherChudzicki: > Note that the `answers` key is provided as a python list for a `ListGrader`, and is a required key.

Re "required": I think what you're getting at here is that multiple input graders cannot get their answers from `answer` or `expect` attributes; author must pass them directly to the grader with keyword arguments.

Do you know if that right? Maybe a statement to that effect goes better in `edx.md`

(The `answers` key is not truly required, e.g., if ListGrader is used inside another ListGrader.)

docs/grading_lists/single_list_grader.md
```diff
@@ -9,7 +9,7 @@ grader = SingleListGrader(
 )
 ```
 
-To receive full points for this problem, a student would enter `cat, dog` or `dog, cat` into the input box. Entering `cat, octopus` or just `cat` will receive half points.
+To receive full points for this problem, a student would enter `cat, dog` or `dog, cat` into the input box. Entering `cat, octopus` or just `cat` will receive half points. The `answers` key is a python list of the individual answers you wish to require, and is a required key for `SingleListGrader`.
```
ChristopherChudzicki: Is this true? I would have expected that 

```xml
<script type="loncapa/python">
from mitxgraders import SingleListGrader
grader = SingleListGrader()
</script>
<customresponse cfn='grader' expect="['cat', 'dog']">
    
</customresponse>
```
since `SingleListGrader` is an `ItemGrader`.

docs/grading_math/formula_grader.md
```diff
@@ -269,6 +269,8 @@ grader = FormulaGrader(
 
 Tolerances are necessary because of numerical roundoff error that lead to small differences in evaluations of algebraically equivalent expressions. Zero tolerance should be used sparingly, perhaps only with integer sampling and answers.
 
+Note that if the answer is exactly `0` (or can be sampled to be exactly `0`, such as when using integer sampling), percentage tolerances don't work (as any percentage of zero is still zero). This means that the student-supplied answer must also evaluate to exactly zero to be graded correctly. Note that answers like `cos(pi/2)` evaluate to approximately `10^(-16)` due to numerical roundoff error. If you want such answers to be graded correctly, make sure to use an absolute tolerance instead of a relative tolerance on such questions.
```
ChristopherChudzicki: 👍

docs/grading_lists/single_list_grader.md
jolyonb: The expect key is passed as a string, whereas `SingleListGrader` requires a list. It doesn't know how to cast that string to a list.

jolyonb:
This has been updated (ever so slightly) in response to your suggestions.

PR #157, opened 2018-12-07
Clarifying equality checks when objects may be arrays

jolyonb:
There are some equality checks in `expressions.py` against strings. These equality checks give rise to warnings when the object being checked against is an array. We need to check if they're strings before doing such a comparison. Note that I'm not sure I caught all of them here.

Here is a minimal working example to see the problem (before this PR):
```python
from mitxgraders import *
grader = MatrixGrader(
    variables=['A'],
    sample_from={'A': RealMatrices()}
)
print(grader('(A-2*A)^2', '(A-2*A)^2'))
```

I'm also not sure how to test this...

### DIFF
--- a/mitxgraders/helpers/calc/expressions.py
+++ b/mitxgraders/helpers/calc/expressions.py
@@ -979,7 +979,7 @@ class MathExpression(object):
         while data:
             # Result contains the current exponent
             working = data.pop()
-            if working == "-":
+            if isinstance(working, str) and working == "-":
                 result = -result
             else:
                 # working is base, result is exponent
@@ -1110,7 +1110,7 @@ class MathExpression(object):
         """
         data = parse_result[:]
         result = data.pop(0)
-        if result == "+":
+        if isinstance(result, str) and result == "+":
             result = data.pop(0)
         while data:
             op = data.pop(0)


ChristopherChudzicki:
> Here is a minimal working example to see the problem (before this PR):

When I run your example (on master branch), I get 
```pycon
>>>  print(grader('(A-2*A)^2', '(A-2*A)^2'))
{'grade_decimal': 1, 'msg': '', 'ok': True}
```

which is what I would have expected.

What are you seeing?

jolyonb:
Finally gotten back to this. It took me a bit to figure out how to generate those warnings...

What I was doing was running against my system python installation in a course (not in the virtual environment). Apparently I have numpy 1.11.1 installed on my system.

The generated warning is:
```
/Users/jolyon/mitx/courses/8.06x/python_lib/mitxgraders/helpers/calc/expressions.py:1113: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if result == "+":
/Users/jolyon/mitx/courses/8.06x/python_lib/mitxgraders/helpers/calc/expressions.py:982: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if working == "-":
{'grade_decimal': 1, 'msg': '', 'ok': True}
```

This seems like something that will need to be fixed for when edX upgrades, so might as well do it now in preparation.

ChristopherChudzicki:
Ah yes, I can reproduce now with numpy 1.11.1. 👍

PR #158, opened 2018-12-07
Introducing flag to suppress matrix error messages

jolyonb:
This PR introduces a flag to `MatrixGrader` that completely suppresses all matrix error messages. This is useful when variables are secretly being used as matrices due to their noncommutativity, but matrix error messages would be confusing to students. Includes code, tests, documentation and course example.

### DIFF
--- a/mitxgraders/formulagrader/matrixgrader.py
+++ b/mitxgraders/formulagrader/matrixgrader.py
@@ -11,7 +11,7 @@ from mitxgraders.formulagrader.formulagrader import FormulaGrader
 from mitxgraders.helpers.validatorfuncs import NonNegative
 from mitxgraders.helpers.calc import MathArray, within_tolerance, identity
 from mitxgraders.helpers.calc.exceptions import (
-    CalcError, MathArrayShapeError as ShapeError)
+    MathArrayShapeError as ShapeError, MathArrayError, DomainError)
 from mitxgraders.helpers.calc.mathfuncs import (
     merge_dicts, ARRAY_ONLY_FUNCTIONS)
 
@@ -55,6 +55,9 @@ class MatrixGrader(FormulaGrader):
                     'shape': Type and shape information about the expected and
                         received objects is revealed.
 
+        suppress_matrix_messages (bool): If True, suppresses all matrix-related
+            error messages from being displayed. Overrides shape_errors=True and
+            is_raised=True. Defaults to False.
     """
 
     # merge_dicts does not mutate the originals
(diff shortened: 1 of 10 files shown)


ChristopherChudzicki:
Do you think `suppress_shape_messages` would be too verbose? Otherwise, this all looks good.

jolyonb:
That sounds reasonable, and more intuitive, given that it's only suppressing matrix-related messages. I'll go and make the change.

jolyonb:
Updated the flag name.

ChristopherChudzicki:
Note that right now `suppress_matrix_messages` does not suppress all matrix-related errors, just the shape errors. In particular:

```python
from mitxgraders import *
grader = MatrixGrader(
    variables=['A'],
    sample_from={
        'A': RealMatrices()
    },
    max_array_dim=0,
    suppress_messages=True, 
)

# below raises MathArrayError "Cannot raise a matrix to non-integer powers."
result = grader('A', 'A^1.3')

# below raises a DomainError
# (DomainError is not an instance of MathArrayError, but this is matrix-related)
result = grader('A', 'sin(A)')

@jolyon Maybe you should catch `MathArrayError` instead of `ShapeError`s? This would still expose DomainError messages, but those are harder to suppress. Also, see #183 for ideas about an alternative approach.

(Note: the idea in #183 works for customizing DomainError messages, too.)

**Comment by jolyonb** [comment]
Hmm. I can definitely catch `MathArrayError`s and suppress those too. Domain errors ... can also be caught, I think... Let me see what I can do.

**Comment by jolyonb** [comment]
Ok, I've had a go at fixing things. Let me know what you think. I'm not terribly happy with the outcome, as it's somewhat kludgy, but it's a quick and dirty way to get noncommuting "scalars". I updated the documentation with a warning that taking functions of these "scalars" will be problematic. I'll go and comment on #183 now.

File: mitxgraders/formulagrader/matrixgrader.py [hunk 1]
```diff
@@ -88,15 +92,27 @@ def check_response(self, answer, student_input, **kwargs):
...
                 return {'ok': False, 'msg': err.message, 'grade_decimal': 0}
         except InputTypeError as err:
-            if self.config['answer_shape_mismatch']['is_raised']:
+            if self.config['suppress_matrix_messages']:
+                return {'ok': False, 'msg': '', 'grade_decimal': 0}
+            elif self.config['answer_shape_mismatch']['is_raised']:
                 raise
             else:
                 return {'ok': False, 'grade_decimal': 0, 'msg': err.message}
+        except (DomainError, MathArrayError) as err:
```
ChristopherChudzicki: This will also suppress some errors that are not matrix-related, like `sin(1, 3)` or `kronecker(5)` (though I'm not particularly worried about inputs like that, and the result is just marking them wrong, which is reasonable). If you're OK with that, I'm happy to merge.

jolyonb: Hmm. That shouldn't be all that hard to fix, honestly. Let me see if I can do so...

jolyonb: Ok, I think I've fixed this. It required cleaning up our exception structure a little though. Take another look?

ChristopherChudzicki: 👍 Nice idea, looks good.

PR #161, opened 2018-12-27
Fixing whitespace parsing

jolyonb:
There's a bug in the parser due to the fact that parsing isn't actually stripping out whitespaces. Here's a simple example that trips things up:

```python
from mitxgraders import *
grader = FormulaGrader(variables=['x_{ab}'])
pp(grader('x_{a b}', 'x_{a b}'))
```

(Note that variables cannot have a space in the name.)

This PR fixes the issue.

### DIFF
--- a/mitxgraders/helpers/calc/expressions.py
+++ b/mitxgraders/helpers/calc/expressions.py
@@ -528,7 +528,7 @@ class MathParser(object):
             return self.cache[cache_key]
 
         try:
-            parsed = self.raw_parse(expression)
+            parsed = self.raw_parse(expression_no_whitespace)
         except ParseException:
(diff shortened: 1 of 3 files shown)


mitxgraders/helpers/calc/expressions.py
```diff
@@ -528,7 +528,7 @@ def parse(self, expression):
             return self.cache[cache_key]
 
         try:
-            parsed = self.raw_parse(expression)
+            parsed = self.raw_parse(expression_no_whitespace)
```
ChristopherChudzicki: expression_no_whitespace  was defined and we just forgot to use it? Good catch :-)

PR #170, opened 2019-02-23
Asciimath nowrap overscripts

### LINKED ISSUE #169: AsciiMath Preprocessor Errors
Similar to the `conj` issue #163, we also have math processor errors when the following are typed in our boxes:

* `hat`
* `dot`
* `vec`
* `ddot`
* `tilde`
* `bar`

There are likely others that I am unaware of; these are just the ones that I thought of off the top of my mind in testing. (ETA: These are actually all of the ones that we care about. There are 4 other entries that are defined using mover, but they're for thinks like overbraces that are unlikely to ever be used in this context.)

Something about our escaping is bad for these quantities too (ETA: Like conj, they are also movers!). Perhaps we should have a list of things that don't get escaped? (and include `conj` on that list)

ChristopherChudzicki:
Resolves #169

### DIFF
--- a/MJxPrep.js
+++ b/MJxPrep.js
@@ -92,15 +92,6 @@ if (window.MJxPrep) {
     eqn = wrapVariables(eqn)
     eqn = wrapFuncCalls(eqn)
 
-    // wrapFuncCalls correctly transforms 'conj(z)', but
-    // wrapFuncCalls turns 'conj(z' into '{:conj:}(' and
-    // wrapVariables turns 'conj' into '{:conj:}', This is problematic...
-    // conj is defined below as a <mover/> symbol. edX version of asciimath
-    // has trouble with <mover/> symbols wrapped directly in non-breaking group.
-    // This hack avoids lonely, non-breaking, <mover/> symbols.
-    // See https://github.com/mitodl/mitx-grading-library/issues/163 for more.
-    eqn = eqn.replace(/\{:conj:\}/g, "{:text(co)text(nj):}");
-
     // Return the preprocessed equation
     return eqn;
   }
--- a/MJxPrep.js
+++ b/MJxPrep.js
@@ -109,7 +109,7 @@ if (window.MJxPrep) {
   function updateMathJax() {
     if (MathJax.InputJax.AsciiMath) {
       // Grab the AsciiMath object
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki:
I forgot to update my local repo. Updated, rebased, and force-pushed. I deleted the old fix from #164

MJxPrep.js
```diff
@@ -371,14 +384,13 @@ if (window.MJxPrep) {
       var funcEnd = bracketCloses === null ? bracketOpens : bracketCloses + 1
 
       var front = eqn.slice(0, funcStart)
-      var middle = '{:' + eqn.slice(funcStart, funcEnd) +  ':}'
-      eqn = front + middle + eqn.slice(funcEnd)
+      var middle = eqn.slice(funcStart, funcEnd)
+      eqn = front + wrapIfNotOverscript(null, middle) + eqn.slice(funcEnd)
```
jolyonb: I think this is unnecessary, as nothing in the set contains parentheses, and this is only called for completed functions, right?

ChristopherChudzicki: `wrapFuncCalls` wraps incomplete function calls:

```js
wrapFuncCalls('cat(') // returns '{:cat:}(
wrapVariables('cat(') // returns 'cat('
```

Now, I'm not sure that wrapping incomplete function calls is actually necessary... I think when we encountered the issues with `x'/2` and `p(x)/2`, we adopted "wrap everything" as a philosophy, so I wrapped incomplete function calls, too. (Overscript issues has made me question that philosophy...)

jolyonb: Ah, the point was more that the old code is equivalent to the new code, unless I'm missing something?

ChristopherChudzicki: I think that the test I added for `wrapFuncCalls` describes how we want it to behave for overscript symbols:

https://github.com/mitodl/mitx-grading-library/pull/170/files#diff-1d7052fc002008ab0813d9d4de53d3b7R191

The middle two cases for that test require this alteration to `wrapFuncCalls`.

jolyonb: Ah, I see now. Very good! Thanks for the clarifications. For posterity, I'll add that `overline()` converts to `{:overline():}`, which is the desired behavior, and works in AsciiMath.

Issue #174, opened 2019-03-06
Vector comparer


jolyonb:
I'd really, really like a vector comparer that can ignore phases and/or normalization. I know we have one that requires an actual matrix, but I think we should be able to do it independently of a matrix...

ChristopherChudzicki:
Sounds good. How about:

1. `vector_parallel_comparer` checks that student's input is a **nonzero** vector parallel to given vector.
2. `vector_phase_comparer` checks that two vectors are parallel and have the same magnitude

And if it was useful, also could have

3. `vector_span_comparer` check that student's vector is in the span of of a bunch of other vectors.

With appropriate tolerances for everything (nonzeroness, smallness of orthogonal projection, ...)

I could do at least 1 and 2 later tonight if you want.w

jolyonb:
I don't have a use for 3 presently, but 1 and 2 would definitely be helpful!

ChristopherChudzicki:
Resolved by #175

PR #175, opened 2019-03-07
Vector comparers

ChristopherChudzicki:
Addresses #174 by implementing `vector_span_comparer` and `vector_phase_comparer`

Notes:
- I decided not to implement `vector_parallel_comparer`; just use `vector_span_comparer` with one vector.
- Comparer functions `comparer_params` currently have no validation. For example, `vector_span_comparer` expects comparer params to be a list of equal-length vectors. If a list of unequal length vectors is provided, a generic "Input cannot be graded" message is shown, although a more informative message is displayed when `debug=True`.

### DIFF
--- a/mitxgraders/comparers/comparers.py
+++ b/mitxgraders/comparers/comparers.py
@@ -5,9 +5,9 @@ Comparer Functions
 ==================
 
 A comparer function must have signature
-`comparer_func(comparer_params_evals, student_eval, utils)`.
-When `FormulaGrader` (or its subclasses) call your custom comparer function,
-`comparer_func`'s argument values are:
+`comparer_func(comparer_params_evals, student_eval, utils)` and should return
+True or False. When `FormulaGrader` (or its subclasses) call your custom
+comparer function, `comparer_func`'s argument values are:
 
 - `comparer_params_evals`: The `comparer_params` list, numerically evaluated
   according to variable and function sampling.
@@ -45,6 +45,7 @@ NOTE: doctests in this module show how the comparer function would be used
 from numbers import Number
 import numpy as np
 from mitxgraders.exceptions import InputTypeError
+from mitxgraders.helpers.calc.mathfuncs import is_nearly_zero
 
 def equality_comparer(comparer_params_evals, student_eval, utils):
     """
@@ -178,3 +179,126 @@ def eigenvector_comparer(comparer_params_evals, student_eval, utils):
         }
 
     return utils.within_tolerance(actual, expected)
+
+def vector_span_comparer(comparer_params_evals, student_eval, utils):
+    """
+    Check whether student's answer is nonzero and in the span of some given
+    vectors.
+
+    comparer_params: A list of vectors
+
+    Usage
+    =====
+
+    >>> from mitxgraders import MatrixGrader
+    >>> grader = MatrixGrader(
+    ...     answers={
+    ...         'comparer_params': [
+    ...             '[3, x, 1 + i]',
+    ...         ],
+    ...         'comparer': vector_span_comparer
+    ...     },
+    ...     variables=['x'],
+    ... )
+    >>> grader(None, '[3, x, 1 + i]')['ok']
+    True
+    >>> grader(None, '[9, 3*x, 3 + 3*i]')['ok']
+    True
+    >>> grader(None, '[9, 3*x, 3 - 3*i]')['ok']
+    False
+
+    Complex scale factors work, too:
+    >>> grader(None, '(4 + 2*i)*[3, x, 1 + i]')['ok']
+    True
+
+    Student input should be nonzero:
+    >>> grader(None, '[0, 0, 0]')['ok']
+    False
+
+    Input shape is validated:
+    >>> grader(None, '5')
+    Traceback (most recent call last):
+    InputTypeError: Expected answer to be a vector, but input is a scalar
+
+    Multiple vectors can be provided:
+    >>> grader = MatrixGrader(
+    ...     answers={
+    ...         'comparer_params': [
+    ...             '[1, 1, 0]',    # v0
+    ...             '[0, 1, 2]'     # v1
+    ...         ],
+    ...         'comparer': vector_span_comparer
+    ...     },
+    ... )
+
+    Above, 2*v0 + 3i*v1:
+    >>> grader(None, '[2, 2 + 3*i, 6*i]')['ok']
+    True
+    """
+
+    # All comparer_param_evals should have same shape, or numpy.linalg.lstsq
+    # will let us know.
+    utils.validate_shape(student_eval, comparer_params_evals[0].shape)
+
+    if utils.within_tolerance(0, np.linalg.norm(student_eval)):
+        return {
+            'ok': False,
+            'grade_decimal': 0,
+            'msg': 'Input should be a nonzero vector.'
+        }
+
+    # Use ordinary least squares to find an approximation to student_eval
+    # that lies within the span of given vectors, then check that the
+    # residual-sum is small in comparison to student input.
+    column_vectors = np.array(comparer_params_evals).transpose()
+    ols = np.linalg.lstsq(column_vectors, student_eval)
+    error = np.sqrt(ols[1])
+
+    # Check that error is nearly zero, using student_eval as a reference
+    # when tolerance is specified as a percentage
+    return is_nearly_zero(error, student_eval, utils.tolerance)
+
+def vector_phase_comparer(comparer_params_evals, student_eval, utils):
+    """
+    Check that student input equals a given input (to within tolerance), up to
+    an overall phase factor.
+
+    comparer_params: [target_vector]
+
+    Usage
+    =====
+
+    >>> from mitxgraders import MatrixGrader
+    >>> grader = MatrixGrader(
+    ...     answers={
+    ...         'comparer_params': [
+    ...             '[1, exp(-i*phi)]',
+    ...         ],
+    ...         'comparer': vector_phase_comparer
+    ...     },
+    ...     variables=['phi'],
+    ... )
+
+    >>> grader(None, '[1, exp(-i*phi)]')['ok']
+    True
+    >>> grader(None, '[exp(i*phi/2), exp(-i*phi/2)]')['ok']
+    True
+    >>> grader(None, '[i, exp(i*(pi/2 - phi))]')['ok']
+    True
+
+    >>> grader(None, '[1, exp(+i*phi)]')['ok']
+    False
+    >>> grader(None, '[2, 2*exp(-i*phi)]')['ok']
+    False
+
+    """
+    # We'll check that student input is in the span as target vector and that
+    # it has the same magnitude
+
+    in_span = vector_span_comparer(comparer_params_evals, student_eval, utils)
+
+    expected_mag = np.linalg.norm(comparer_params_evals[0])
+    student_mag = np.linalg.norm(student_eval)
+    same_magnitude = utils.within_tolerance(expected_mag, student_mag)
+
+    return in_span and same_magnitude
--- a/mitxgraders/comparers/comparers.py
+++ b/mitxgraders/comparers/comparers.py
@@ -242,7 +242,7 @@ def vector_span_comparer(comparer_params_evals, student_eval, utils):
     >>> grader(None, '[2, 2 + 3*i, 6*i]')['ok']
     True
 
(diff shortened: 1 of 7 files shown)


mitxgraders/comparers/comparers.py
```diff
@@ -178,3 +179,126 @@ def eigenvector_comparer(comparer_params_evals, student_eval, utils):
...
+    >>> grader(None, '[2, 2 + 3*i, 6*i]')['ok']
+    True
+    """
+
+    # All comparer_param_evals should have same shape, or numpy.linalg.lstsq
+    # will let us know.
+    utils.validate_shape(student_eval, comparer_params_evals[0].shape)
+
+    if utils.within_tolerance(0, np.linalg.norm(student_eval)):
+        return {
```
jolyonb: Shouldn't this just return False?

jolyonb:
Travis failed on a node.js thing, so I tried running it again.

Otherwise, I just saw the one thing that jumped out at me. I'm going to take another look when I'm awake however!

mitxgraders/comparers/comparers.py
ChristopherChudzicki: You mean instead of the `{ok, msg, grade}` dict? If I remember correctly comparer function is similar to edx check function and can return either True/False or a `{ok, msg, grade}` dict.

(Although, I just realized I updated the comment at top of file to say only True/False. That was incorrect.)

ChristopherChudzicki:
Yeah, I'm not sure why Travis is failing on node tests. That's weird, since I didn't change anything js-related.

I'm going to update package-lock file now. If that doesn't fix it, I'll look into it more tomorrow... My local node and npm versions are slightly different than Travis's.

mitxgraders/comparers/comparers.py
jolyonb: I think there's no particular reason to return the dict here, so just return False. You are correct though that you can return the dict.

jolyonb: Ah, ignore me - I see that you're returning a message :-)

ChristopherChudzicki:
@jolyon  Node tests passing now 🎊.

We tell Travis to use the latest version of Node. Since our last working Travis build, latest changed from `11.10.0` to `11.11.0`, which broke `jest` (our node test runner). Upgrading `jest` fixed the issue.

mitxgraders/comparers/comparers.py
```diff
@@ -178,3 +180,131 @@ def eigenvector_comparer(comparer_params_evals, student_eval, utils):
...
+    >>> grader(None, '[1, exp(+i*phi)]')['ok']
+    False
+    >>> grader(None, '[2, 2*exp(-i*phi)]')['ok']
+    False
+
+    """
+    # We'll check that student input is in the span as target vector and that
+    # it has the same magnitude
+
+    in_span = vector_span_comparer(comparer_params_evals, student_eval, utils)
```
jolyonb: Can we add a check here that `comparer_params_evals` has a single entry and is a vector?

mitxgraders/comparers/comparers.py
```diff
@@ -178,3 +180,131 @@ def eigenvector_comparer(comparer_params_evals, student_eval, utils):
...
+    ...         'comparer': vector_span_comparer
+    ...     },
+    ... )
+
+    The vector 2*v0 + 3i*v1 = [2, 2+3i, 6i] is in the span of v0 and v1:
+    >>> grader(None, '[2, 2 + 3*i, 6*i]')['ok']
+    True
+    """
+
+    # All comparer_param_evals should have same shape, or numpy.linalg.lstsq
```
jolyonb: Can we add a check here that all `comparer_param_evals` are vectors and all have the same shape?

mitxgraders/helpers/calc/mathfuncs.py
```diff
@@ -380,10 +398,44 @@ def within_tolerance(x, y, tolerance):
...
 
     return np.linalg.norm(difference) <= tolerance
+
+def is_nearly_zero(x, reference, tolerance):
+    """
+    Check that x is nearly zero in comparison to reference.
+
+    Args:
+        x: number or array (np array_like)
+        reference: number or array (np array_like)
```
jolyonb: Note that `reference` is only needed when using a percentage tolerance.

mitxgraders/helpers/calc/mathfuncs.py
```diff
@@ -380,10 +398,44 @@ def within_tolerance(x, y, tolerance):
...
-        tolerance = np.linalg.norm(x) * float(tolerance[:-1]) * 0.01
+        tolerance = np.linalg.norm(x) * percentage_as_number(tolerance)
 
     difference = x - y
 
     return np.linalg.norm(difference) <= tolerance
+
+def is_nearly_zero(x, reference, tolerance):
+    """
+    Check that x is nearly zero in comparison to reference.
```
jolyonb: Not strictly true; in comparison to tolerance. Maybe make reference have a default value of None?

mitxgraders/comparers/comparers.py
```diff
@@ -178,3 +180,131 @@ def eigenvector_comparer(comparer_params_evals, student_eval, utils):
...
+            'ok': False,
+            'grade_decimal': 0,
+            'msg': 'Input should be a nonzero vector.'
+        }
+
+    # Use ordinary least squares to find an approximation to student_eval
+    # that lies within the span of given vectors, then check that the
+    # residual-sum is small in comparison to student input.
+    column_vectors = np.array(comparer_params_evals).transpose()
+    ols = np.linalg.lstsq(column_vectors, student_eval)
```
jolyonb: When we move to a newer version of numpy, we might want to add `rcond=None` to this.

jolyonb:
Yay on Travis working again :-)

I've gone through the code in detail now, and added some more minor comments. Everything else looks great!

mitxgraders/comparers/comparers.py
ChristopherChudzicki: I've added this sort of validation for `vector_span_comparer` and `vector_phase_comparer`. Two downsides:

1. Will run for each sample during grading
2. Raises `StudentFacingError`, since validation happens upon "check" not upon instantiation.

ChristopherChudzicki: (Minor downsides, not worth worrying about right now. Overcoming them would require validating comparer_params during instantiation, which would be a lot of refactoring. We have basically the same issue for validating author answers with default comparer.)

mitxgraders/helpers/calc/mathfuncs.py
ChristopherChudzicki: Added a note.

mitxgraders/comparers/comparers.py
jolyonb: I agree - given that we don't validate author answers until check (and to be fair, neither does edX!), these are indeed minor downsides.

mitxgraders/comparers/comparers.py
```diff
@@ -240,10 +241,28 @@ def vector_span_comparer(comparer_params_evals, student_eval, utils):
     The vector 2*v0 + 3i*v1 = [2, 2+3i, 6i] is in the span of v0 and v1:
     >>> grader(None, '[2, 2 + 3*i, 6*i]')['ok']
     True
+
+    The compaer_params should be list of equal-length vectors:
```
jolyonb: Typo in `comparer_params`

mitxgraders/comparers/comparers.py
```diff
@@ -240,10 +241,28 @@ def vector_span_comparer(comparer_params_evals, student_eval, utils):
...
+    >>> grader(None, '[1, 2, 3]')               # doctest: +ELLIPSIS
+    Traceback (most recent call last):
+    StudentFacingError: Problem Configuration Error: ...to equal-length vectors
     """
 
-    # All comparer_param_evals should have same shape, or numpy.linalg.lstsq
-    # will let us know.
+    # Validate the comparer params
+    if not are_same_length_vectors(comparer_params_evals):
+        raise StudentFacingError('Problem Configuration Error: comparer_params'
```
jolyonb: Needs a space at the end, I think?

mitxgraders/comparers/comparers.py
```diff
@@ -257,12 +276,14 @@ def vector_span_comparer(comparer_params_evals, student_eval, utils):
     # that lies within the span of given vectors, then check that the
     # residual-sum is small in comparison to student input.
     column_vectors = np.array(comparer_params_evals).transpose()
-    ols = np.linalg.lstsq(column_vectors, student_eval)
+    # rcond=-1 uses machine precision for testing singular values
+    # In numpy 1.14+, use rcond=None fo this behavio. (we use 1.6)
```
jolyonb: Typo in `behavior`

mitxgraders/comparers/comparers.py
```diff
@@ -297,7 +318,25 @@ def vector_phase_comparer(comparer_params_evals, student_eval, utils):
     >>> grader(None, '[2, 2*exp(-i*phi)]')['ok']
     False
 
+    The compaer_params should be list of equal-length vectors:
```
jolyonb: Typo in `comparer_params`

jolyonb: Also, should be a list containing a single vector, not a list of equal length vectors.

mitxgraders/helpers/calc/mathfuncs.py
```diff
@@ -404,38 +404,48 @@ def within_tolerance(x, y, tolerance):
...
+    >>> is_nearly_zero(x, '5%', reference=10)
     True
-    >>> is_nearly_zero(x, 10, '1.5%')
+    >>> is_nearly_zero(x, '1.5%', reference=10)
     False
+
+    A ValueError is raised when percentage tolerance is used without reference:
+    >>> is_nearly_zero(0.4, '3%')
+    Traceback (most recent call last):
+    ValueError: When tolerance is a percentage, refeence must not be None.
```
jolyonb: Typo in `reference`

mitxgraders/helpers/calc/mathfuncs.py
```diff
@@ -404,38 +404,48 @@ def within_tolerance(x, y, tolerance):
...
     """
     # When used within graders, tolerance has already been
     # validated as a Number or PercentageString
     if isinstance(tolerance, str):
+        if reference is None:
+            raise ValueError('When tolerance is a percentage, refeence must '
```
jolyonb: Typo in `reference`

jolyonb:
Love the code. I think every comment I just made had nothing to do with the actual code!

mitxgraders/comparers/comparers.py
ChristopherChudzicki: So I honestly can't remember why I wrote `rcond=-1`. That seems to be the default for our version of numpy anyway:

https://docs.scipy.org/doc/numpy-1.6.0/reference/generated/numpy.linalg.lstsq.html

The new default value (np 1.14+) should be only slightly bigge for our low-dimensional matrices:

https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.lstsq.html

@jolyonb Maybe we could just let numpy do it's defaults?

With unspecified `rcond`, my (fairly minimal) tests pass.

mitxgraders/comparers/comparers.py
ChristopherChudzicki: So, the `r` button on my keyboard only works 30% of the time :) Fixing that soon I hope...

mitxgraders/comparers/comparers.py
ChristopherChudzicki: Thanks, fixed.

mitxgraders/comparers/comparers.py
jolyonb: I'm happy to leave this argument off for the moment, as things seem to work. We're expecting edX to upgrade to python 3.6 and the latest version of numpy+scipy later this year, so just flagging this as something that may need attention. Perhaps just leave a comment?

mitxgraders/comparers/comparers.py
jolyonb: Man, that would dive me cazy!

ChristopherChudzicki:
Thanks, I think I fixed all the missing `r`s. Getting my keyboard fixed next week :) 

Merge when want. Hope it helps with 8.06!

jolyonb:
Looks great! I'll add documentation to my todo list.

Issue #171, opened 2019-02-23
Asciimath: reserved names


ChristopherChudzicki:
@jolyonb In Issue #169 discussion, you mentioned:

> I think we should add to the documentation that there are a list of names that are reserved and should not be used due to asciimath, and include all those movers. Can you imagine a variable called `hat'`

I just made PR #170 to resolve issue #169, but decided not to address the reserved name documentation issue.

The list of "reserved names" is pretty big, right? It includes things like:

- `int` (integral sign)
- `in` (element of)
- `ne` (not equal)
- `le` (less than or equal)

so `fin` displays as `f∈`

In my opinion, it's not worth documenting this. At most, we should say "`<textline />` includes a whole bunch of extra symbols compared to `<formulaequationinput/>`. Make sure that your problem displays correctly before releasing".

jolyonb:
I think you make a good point. There's far too much stuff in asciimath to state it all, and the fact that we pollute the namespace a tiny amount more isn't worth documenting in detail. A simple statement that one should check how things display before using them should suffice.

jolyonb:
Resolved by #180

PR #180, opened 2019-03-12
Documenting variable names

jolyonb:
Addresses #171 by adding documentation about possible variable names (not sure why this was previously completely omitted! Either that, or I'm blind...).


jolyonb:
Note that we'll need to run `tidy.sh` after merging all of these documentation updates...

Issue #159, opened 2018-12-07
Matrix sampler - identity * constant


jolyonb:
I need this sampler for something that commutes with other matrices properly...

jolyonb:
Resolved by #182

PR #182, opened 2019-03-19
Adding IdentityMatrixMultiples sampler

jolyonb:
This creates a sampler to sample matrices that are scalar multiples of the identity matrix. It determines that scalar by accepting another sampler. To do this, I added an extra abstract subclass of `VariableSamplingSet` called `ScalarSamplingSet`. I also created a new abstract class `AbstractSquareMatrices` that can be used to build up various special types of matrices. Documentation is included.

### DIFF
--- a/mitxgraders/sampling.py
+++ b/mitxgraders/sampling.py
@@ -222,7 +222,7 @@ class ComplexSector(ScalarSamplingSet):
         return self.modulus.gen_sample() * np.exp(1j * self.argument.gen_sample())
 
 
-class DiscreteSet(ScalarSamplingSet):  # pylint: disable=too-few-public-methods
+class DiscreteSet(VariableSamplingSet):  # pylint: disable=too-few-public-methods
     """
     Represents a discrete set of values from which to sample.
 
--- a/mitxgraders/sampling.py
+++ b/mitxgraders/sampling.py
@@ -392,7 +392,6 @@ class IdentityMatrixMultiples(AbstractSquareMatrices):
 
     The scalar multiple can be generated in a number of ways:
     >>> matrices = IdentityMatrixMultiples(sampler=[1,3])
-    >>> matrices = IdentityMatrixMultiples(sampler=(1, 3, 5))
     >>> sect = ComplexSector(modulus=[0,1], argument=[-np.pi,np.pi])
     >>> matrices = IdentityMatrixMultiples(sampler=sect)
 
--- a/mitxgraders/sampling.py
+++ b/mitxgraders/sampling.py
@@ -1,14 +1,18 @@
 """
 sampling.py
 
-Contains classes for sampling numerical values
+Contains classes for sampling numerical values:
 * RealInterval
 * IntegerRange
 * DiscreteSet
 * ComplexRectangle
 * ComplexSector
 * DependentSampler
-and for specifying functions
+for sampling vectors/matrices/tensors:
+* RealMatrices
+* RealVectors
+* IdentityMatrixMultiples
+for specifying functions:
 * SpecificFunctions
 * RandomFunction
 
@@ -43,6 +47,7 @@ __all__ = [
     "ComplexSector",
     "RealVectors",
     "RealMatrices",
+    "IdentityMatrixMultiples",
     "SpecificFunctions",
     "RandomFunction",
     "DependentSampler"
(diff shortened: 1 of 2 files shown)


jolyonb:
Resolves #159.

mitxgraders/sampling.py
```diff
@@ -338,6 +349,85 @@ class RealMatrices(RealMathArrays):
...
+    def gen_sample(self):
+        """
+        Generates an identity matrix of specified dimension multiplied by a random scalar
+        """
+        # Sample the multiplicative constant
+        scaling = self.config['sampler'].gen_sample()
+        # Create the matrix
+        matrix = scaling * np.eye(self.config['dimension'])
+        # Return the result
+        return matrix
```
ChristopherChudzicki: This needs to be wrapped in `MathArray` (and then probably need to fix some doctests above. 

Normal numpy arrays can't be used in MatrixGrader:

```python
from mitxgraders import *
grader = MatrixGrader(
    variables=['A', 'D'],
    sample_from={
        'A': RealMatrices(),
        'D': IdentityMatrixMultiples()
    },
    debug=True
)

# TypeError: Cannot multiply a matrix with object of <type 'numpy.ndarray'>.
result = grader('D*A', 'A*D')
```

mitxgraders/sampling.py
```diff
@@ -338,6 +349,85 @@ class RealMatrices(RealMathArrays):
...
+    })
+
+    def gen_sample(self):
+        """
+        Generates an identity matrix of specified dimension multiplied by a random scalar
+        """
+        # Sample the multiplicative constant
+        scaling = self.config['sampler'].gen_sample()
+        # Create the matrix
+        matrix = scaling * np.eye(self.config['dimension'])
```
ChristopherChudzicki: Suggestion: rename this `array` or something, since it's not a MathArray Matrix.

mitxgraders/sampling.py
```diff
@@ -210,12 +222,12 @@ def gen_sample(self):
         return self.modulus.gen_sample() * np.exp(1j * self.argument.gen_sample())
 
 
-class DiscreteSet(VariableSamplingSet):  # pylint: disable=too-few-public-methods
+class DiscreteSet(ScalarSamplingSet):  # pylint: disable=too-few-public-methods
```
ChristopherChudzicki: Prior to this PR, `DiscreteSet` could sample from a list of matrices, right? 

I understand that this change lets `DiscreteSet` be used as a sampler for `IdentityMatrixMultiples`. But if you need this feature, it seems better to me to keep DiscreteSet general (since its name sounds general) and have a specialized DiscreteScalarSet. 

Or, if you don't need this feature, just remove DiscreteSet as one of the possible IdentityMatrixMultiples samplers, and we can cross this bridge when necessary.

mitxgraders/sampling.py
jolyonb: Ok, can do.

mitxgraders/sampling.py
jolyonb: Ah, good call. Fixing now.

mitxgraders/sampling.py
jolyonb: It actually can't. `DiscreteSet` is defined using the schema `schema_config = Schema(TupleOfType(Number))`, so if we want to be able to have matrices here, we need to update `DiscreteSet`. However, I think your point is valid: having matrices in `DiscreteSet` obviates the need to be able to use `DiscreteSet` in `IdentityMatrixSampler` (just pass in the appropriate matrices to `DiscreteSet`!)

jolyonb:
I've made #184 to handle the `DiscreteSet` issue. Otherwise, I think this one is all cleaned up and ready.

mitxgraders/sampling.py
```diff
@@ -338,6 +349,83 @@ class RealMatrices(RealMathArrays):
...
+    (2, 2)
+
+    We can generate NxN matrices by specifying the dimension:
+    >>> matrices = IdentityMatrixMultiples(dimension=4)
+    >>> matrices.gen_sample().shape
+    (4, 4)
+
+    The scalar multiple can be generated in a number of ways:
+    >>> matrices = IdentityMatrixMultiples(sampler=[1,3])
+    >>> matrices = IdentityMatrixMultiples(sampler=(1, 3, 5))
```
ChristopherChudzicki: I'm going to remove this line and merge—this looks like old test before you removed DiscreteSet as asampler

jolyonb: Ah, my bad. Thanks!

Issue #83, opened 2018-07-13
Matrix samplers


jolyonb:
There are a number of special types of matrices that we might want to sample:

Properties:
* (Anti-)Symmetric
* Traceless
* (Anti-)Hermitian
* (Non-)Zero Determinant or Unit Determinant
* Diagonal

Special kinds:
* Orthogonal O(n) or SO(n) (see [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.ortho_group.html) and [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.special_ortho_group.html)) (implemented in #186 but blocked due to edX)
* Unitary U(n) or SU(n)  (see [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.unitary_group.html)) (implemented in #186 but blocked due to edX)
* Upper/Lower Triangular (implemented in #187)

I'm sure there are many, many more possibilities, but the above should keep most people out of trouble! I suspect that most of these are pretty easy to do, too.

jolyonb:
A lot of these should be straightforward to do, building on the template given in #182.

jolyonb:
Just thinking a bit though, it might be possible to do some of these in a really neat manner. Maybe make a generic `SquareMatrixSampler` that has a `properties` list in the schema, with possible terms including `antisymmetric`, `symmetric`, `traceless`, `detzero`, `detone` and `diagonal`. (We should allow for a given norm, and real or complex entries too.) While these terms need to be able to "mix and match", the rest of the matrix types listed shouldn't need to have these properties specified.

We should probably also add complex matrix sampler versions of the `RealMatrices` and `RealVectors` samplers.

jolyonb:
So, thinking about this a little more, `SquareMatrixSampler` could have the following properties:

* `dimension`: (integer, preferably >= 2)
* `type`: `real` or `complex`
* `symmetry`: `any`, `diagonal`, `symmetric`, `antisymmetric`, `hermitian`, `antihermitian`
* `trace`: `any`, `zero`
* `det`: `any`, `one`, `zero`
* `norm`: sampling set for norm, ignored if `det` is `one`.

[...]

PR #187, opened 2019-03-22
Adding triangular matrix sampling

jolyonb:
This implements triangular matrix sampling, as described in #83. Documentation and tests included.

### DIFF
--- a/docs/grading_math/sampling.md
+++ b/docs/grading_math/sampling.md
@@ -96,6 +96,20 @@ Sample real matrices of a specific shape and norm. (`RealMatrices` uses the Frob
 
 If you want to sample a matrix in a way that depends on a scalar, see `DependentSampler` below.
 
+### TriangularMatrices
+
+Sample triangular matrices of a specific shape and norm.
+
+```python
+>>> # Sample lower triangular 3 by 2 real matrices with norm between 5 and 10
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki:
Nice when we can make numpy do the work 😄 

Two questions:

First, are you thinking of
    - individual classes (Triangle, Orthogonal, Unitary...), or 
    - the generic version `SquareMatrixSampler` described in #83 with mixable keywords, or
    - both?

If both, will we need a lot of duplicate code? I haven't though too much about it.

Second: If separate classes, should `TriangularMatrices ` be real or complex? (Not really an issue if we make `SquareMatrixSampler` the main author-facing API.)

jolyonb:
Well, triangular is special, because it doesn't just apply to square matrices. Also, unitary and orthogonal are also special, because you can't specify anything other than the dimension and determinant. Everything else can be covered by a `SquareMatrices` class, I think. However, this probably requires a little more thought as to how our matrix classes should be constructed... I'll make an issue.

ChristopherChudzicki:
This still needs review and merge, right? Sorry it fell through the cracks, I was on vacation in March.

jolyonb:
I think at this stage, this type of thing should be incorporated into the more general matrix sampling scheme that I outlined in #188. I've left this here for the moment though.

jolyonb:
Closing this in favor of an upcoming PR.

PR #189, opened 2019-04-11
fix bug partial credit for multi-answer ListGrader

ChristopherChudzicki:
Fixes a bug where ListGraders was having trouble assigning partial credit to problems where `answers` was specified as a tuple of answers.

**Note:** The bugfix is really just transposing one matrix to fix a shape comparison. I believe the bug has gone undetected until now because our test case used 2 inputs and had 2 author-supplied answer, and transpose of a 2x2 is 2x2.

@jolyonb I believe this fixes the issue your 8.06 student encountered. I really would like to go over the code more carefully, add comments, etc, but you have an exam running and I'm about to get off my bus!  I did add a more descriptive docstring.

### DIFF
--- a/mitxgraders/listgrader.py
+++ b/mitxgraders/listgrader.py
@@ -510,7 +510,18 @@ class ListGrader(AbstractGrader):
 
     @staticmethod
     def get_best_result(results):
-        """Compute the best result from a multi-input problem"""
+        """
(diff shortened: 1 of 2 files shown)


jolyonb:
This looks great. Thanks so much for tracking this down!!!

PR #196, opened 2019-06-22
Correlated comparers

ChristopherChudzicki:
This is a re-worked version of #194. (The first several commits laying groundwork for correlated comparers are the same.)

Implements configurable AffineComparer.

Introduces `FormulaGrader.set_default_comparer(comparer)` and `FormulaGrader.reset_default_comparer()` methods. (See docstrings about why not to set the `default_comparer` method directly.)


ChristopherChudzicki:
Rebased to clean up some commits; should have done that first 🙍‍♂

ChristopherChudzicki:
@jolyonb This is ready for review. I believe I've implemented what we talked about.

BTW: I think we should rename `AffineComparer` to `LinearComparer`. Linear is a common word for `y=ax+b`, and since we're using "proportional" (not linear) for `y=ax`, I don't think there's a conflict. But in case you want to keep it, I did not make this change yet.

mitxgraders/comparers/affine_comparer.py
```diff
@@ -0,0 +1,167 @@
...
+    >>> x = np.array([2, 5, 8])
+    >>> result = get_affine_fit_error(x, 2*x + 1)
+    >>> round(result, 8)
+    0.0
+    """
+    A = np.vstack([x, np.ones(len(x))]).T
+    _, residuals, _, _ = np.linalg.lstsq(A, y, rcond=-1)
+    return np.sqrt(residuals)
+
+def get_proportional_fit_error(x, y):
```
ChristopherChudzicki: @jolyonb I decided to essentially do 4 separate fits to estimate the error. The `get_affine_error` function fits `y=ax+b`, and the `get_proportional_error` function fits `y=ax`. (The other two never actually do the fits, but still estimate the error.)

This seemed easier than fitting `y=ax+b` once and trying to extract meaning for the 4 separate cases from the coefficients and residuals.

mitxgraders/comparers/affine_comparer.py
```diff
@@ -0,0 +1,167 @@
...
+    mean = np.mean(y - x)
+    return np.sqrt(sum(np.square(x + mean - y)))
+
+def get_equals_fit_error(x, y):
+    """
+    Get total error in the difference between two samples.
+    Arguments:
+        x, y: compatible numpy arrays
+    """
+    return sum(np.abs(x-y))
```
jolyonb: May I suggest doing this RMS like you do the others?

mitxgraders/comparers/affine_comparer.py
```diff
@@ -0,0 +1,167 @@
...
+        equals (None | number): defaults to 1.0
+        proportional (None | number): defaults to 0.5
+        offset (None | number): defaults to None
+        affine (None | number): defaults to None
+
+    The remaining configuration keys specify a feedback message to be given
+    in each case:
+        equals_msg (str): defaults to ''
+        proportional_msg (str): defaults to 'The submitted answer differs from
+            an expected answer by a constant factor.'
```
jolyonb: "an expected answer" is kind of awful, but I see why you went with it. Hrmm...

jolyonb:
I've taken a quick look (still need to do line-by-line). Things look pretty good!

I can approve the change from affine -> linear. It's funny - a linear transformation is x = My, whereas an affine transformation is x = My + b. However, a linear relationship is x = My + b. Silliness.

mitxgraders/comparers/affine_comparer.py
ChristopherChudzicki: done

mitxgraders/comparers/linear_comparer.py
```diff
@@ -0,0 +1,169 @@
...
+
+    The remaining configuration keys specify a feedback message to be given
+    in each case:
+        equals_msg (str): defaults to ''
+        proportional_msg (str): defaults to 'The submitted answer differs from
+            an expected answer by a constant factor.'
+        offset_msg (str): defaults to ''
+        linear_msg (str): defaults to ''
+
+    NOTE:
```
jolyonb: What's this note?

mitxgraders/comparers/linear_comparer.py
```diff
@@ -0,0 +1,169 @@
...
+        'proportional': get_proportional_fit_error,
+        'offset': get_offset_fit_error,
+        'linear': get_linear_fit_error,
+    }
+
+    def __call__(self, comparer_params_evals, student_evals, utils):
+        student_eval_norm = np.linalg.norm(student_evals)
+
+        # Validate student input shape...only needed for MatrixGrader
+        try:
```
jolyonb: Any reason to not do the same thing `EqualityComparer` does and check for the `validate_shape` attribute?

mitxgraders/comparers/linear_comparer.py
```diff
@@ -0,0 +1,169 @@
...
+            utils.validate_shape(student_evals[0], comparer_params_evals[0][0].shape)
+        except AttributeError:
+            pass # not called by MatrixGrader
+
+        if is_nearly_zero(student_eval_norm, utils.tolerance, reference=comparer_params_evals):
```
jolyonb: Uhh... the reference here is off. `comparer_params_evals` is a list of evaluations. Also, `student_eval_norm` is a single number; shouldn't this comparison be done for each input individually? Finally, I think putting this statement here precludes checking for equality when the answer is actually 0.

mitxgraders/comparers/linear_comparer.py
```diff
@@ -0,0 +1,169 @@
...
+            return False
+
+        # Get the result for each mode
+        student = np.array(student_evals).flatten()
```
jolyonb: Add a comment here that the flatten is to handle array inputs.

mitxgraders/comparers/linear_comparer.py
```diff
@@ -0,0 +1,169 @@
...
+        expected = np.array(comparer_params_evals).flatten()
+        errors = [self.error_calculators[mode](student, expected) for mode in self.modes]
+        results = [
+            {'grade_decimal': self.config[mode], 'msg': self.config[mode+'_msg']}
+            if is_nearly_zero(error, utils.tolerance, reference=student_eval_norm)
+            else
+            {'grade_decimal': 0, 'msg': ''}
+            for mode, error in zip(self.modes, errors)
```
jolyonb: This is a bit unusual - the comparer is setting the `msg`. It's also possible for the answer to have a `msg` setting associated with it. I'm just noting this here because we should probably make sure that any message set here isn't overwritten? I'll read on...

mitxgraders/formulagrader/formulagrader.py
```diff
@@ -391,12 +391,17 @@ class FormulaGrader(ItemGrader):
...
 
-        answers (str | dict): A string, dictionary, or tuple thereof. If a string is supplied,
-            it represents the correct answer and is compared to student input for equality.
+        answers: A single "expect" value, a dictionary, or a tuple thereof, as
+            described in the documentation for ItemGraders.
 
-            If a dictionary is supplied, it needs keys:
+            The expect value can be a string, or can itself be a dictionary.
+
+            If the expect value is a string string, it represents the correct
```
jolyonb: `string` is repeated twice.

mitxgraders/formulagrader/formulagrader.py
```diff
@@ -409,6 +414,32 @@ class FormulaGrader(ItemGrader):
     default_variables = DEFAULT_VARIABLES.copy()
     default_suffixes = DEFAULT_SUFFIXES.copy()
 
+    default_comparer = staticmethod(equality_comparer)
```
jolyonb: Huh, never seen that done before!

mitxgraders/formulagrader/formulagrader.py
```diff
@@ -696,34 +744,104 @@ def scoped_eval(expression,
...
+        """
+
+        # answer contains extra keys, so prune them
+        pruned_answer = { key: answer[key] for key in ['ok', 'grade_decimal', 'msg'] }
+
+        correlated = isinstance(answer['expect']['comparer'], CorrelatedComparer)
+
+        # Correlated comparers return a single result, so failable_evals makes no sense
+        if correlated:
+            [result] = results
```
jolyonb: Can we instead do `result = results[0]` in this case? It has the same effect, but it's easier to understand.

mitxgraders/formulagrader/formulagrader.py
```diff
@@ -696,34 +744,104 @@ def scoped_eval(expression,
...
+        # Get the comparer function
+        comparer = answer['expect']['comparer']
+        results = self.compare_evaluations(comparer_params_evals, student_evals,
+                                           comparer, self.comparer_utils)
+        # Comparer function results might assign partial credit.
+        # But the answer we're testing against might only merit partial credit
+        for result in results:
+            result['grade_decimal'] *= answer['grade_decimal']
+        consolidated = self.consolidate_results(results, answer, self.config['failable_evals'])
+
```
jolyonb: Linting - extra spaces.

mitxgraders/formulagrader/formulagrader.py
```diff
@@ -696,34 +744,104 @@ def scoped_eval(expression,
...
+        """
+        results = []
+        if isinstance(comparer, CorrelatedComparer):
+            result = comparer(compare_parms_evals, student_evals, utils)
+            results.append(ItemGrader.standardize_cfn_return(result))
+        else:
+            for compare_parms_eval, student_eval in zip(compare_parms_evals, student_evals):
+                result = comparer(compare_parms_eval, student_eval, utils)
+                results.append(ItemGrader.standardize_cfn_return(result))
+
```
jolyonb: Linting - extra line.

mitxgraders/formulagrader/formulagrader.py
```diff
@@ -696,34 +744,104 @@ def scoped_eval(expression,
...
+        Consolidate comparer result(s) into just one result.
+
+        Arguments:
+            results: a list of results dicts
+            answer (dict): correctness data for the expected answer
+            failable_evals: int
+        """
+
+        # answer contains extra keys, so prune them
+        pruned_answer = { key: answer[key] for key in ['ok', 'grade_decimal', 'msg'] }
```
jolyonb: Whenever `pruned_answer` is returned, it ignores any `msg` entries that the comparer may have generated. This is probably ok, but we should document that `msg` entries from comparers are ignored if the comparer finds agreement. What do you think?

mitxgraders/comparers/linear_comparer.py
jolyonb: Ok, looks like the `msg` from a comparer is ignored if the comparer evaluates true. See comment below.

mitxgraders/comparers/affine_comparer.py
jolyonb: Yeah, I think this was a good call.

mitxgraders/comparers/linear_comparer.py
ChristopherChudzicki: Adding a note about matrix/vectors.

mitxgraders/comparers/linear_comparer.py
ChristopherChudzicki: Done

mitxgraders/comparers/linear_comparer.py
ChristopherChudzicki: done

mitxgraders/formulagrader/formulagrader.py
ChristopherChudzicki: That's all that `@decorator` does 😄

mitxgraders/formulagrader/formulagrader.py
ChristopherChudzicki: Recently I've been preferring `[result] = results` over `result = results[0]` if I expect `results` to have length 1. The first first version will throw a `ValueError:  too many values to unpack` if `results` is longer than it should be, the second will not.

jolyonb: Hmm, ok.

mitxgraders/formulagrader/formulagrader.py
ChristopherChudzicki: Yes, let's document it. I do think this is the correct behavior...I was tempted to concatenate the two messages, but just because a comparer returns `'ok': True` doesn't mean the input is correct (i.e., comparing could be checking one of several `answers=...` partial credit values).

Really, all error messages from comparers suffer this problem, hence my terrible "The submitted answer differs from an expected answer by a constant factor.".

jolyonb: Perhaps this belongs in the general discussion regarding comparers then. It seems to be an appropriate policy.

jolyonb: (I do realize that we aren't writing docs in this PR, but perhaps add it to a docstring?)

mitxgraders/comparers/linear_comparer.py
```diff
@@ -0,0 +1,203 @@
...
+    @staticmethod
+    def check_comparing_zero(comparer_params_evals, student_evals, tolerance):
+        """
+        Check whether student input is nearly zero, or author input is exactly zero
+        """
+        student_zero = all([
+            is_nearly_zero(x, tolerance, reference=y)
+            for x, y in zip(comparer_params_evals, student_evals)
+        ])
+        expected_zero = all(x == 0.0 for [x] in comparer_params_evals)
```
ChristopherChudzicki: @jolyon This change was in response to 
> Uhh... the reference here is off. comparer_params_evals is a list of evaluations. Also, student_eval_norm is a single number; shouldn't this comparison be done for each input individually? Finally, I think putting this statement here precludes checking for equality when the answer is actually 0.

I checked for 0.0 instead of is_nearly_zero for author input since I'm not sure what reference value. Definitely option to suggestions here.

mitxgraders/comparers/linear_comparer.py
```diff
@@ -139,28 +147,54 @@ def __init__(self, config=None, **kwargs):
...
+        # flatten in case individual evals are arrays (as in MatrixGrader)
         student = np.array(student_evals).flatten()
         expected = np.array(comparer_params_evals).flatten()
-        errors = [self.error_calculators[mode](student, expected) for mode in self.modes]
+        errors = [self.error_calculators[mode](student, expected) for mode in filtered_modes]
+
         results = [
             {'grade_decimal': self.config[mode], 'msg': self.config[mode+'_msg']}
-            if is_nearly_zero(error, utils.tolerance, reference=student_eval_norm)
+            if is_nearly_zero(error, utils. tolerance, reference=student_evals_norm)
```
jolyonb: There's an extra space here that looks like a typo?

mitxgraders/comparers/linear_comparer.py
```diff
@@ -139,28 +147,54 @@ def __init__(self, config=None, **kwargs):
         'linear': get_linear_fit_error,
     }
 
+    @staticmethod
+    def check_comparing_zero(comparer_params_evals, student_evals, tolerance):
+        """
+        Check whether student input is nearly zero, or author input is exactly zero
+        """
+        student_zero = all([
+            is_nearly_zero(x, tolerance, reference=y)
```
jolyonb: Do you have x and y switched here?

mitxgraders/comparers/linear_comparer.py
jolyonb: I think this is fine.

mitxgraders/comparers/linear_comparer.py
```diff
@@ -139,28 +147,54 @@ def __init__(self, config=None, **kwargs):
...
+
+
+    def get_valid_modes(self, is_comparing_zero):
+        """
+        Returns a copy of self.modes, first removing 'proportional' and 'linear'
+        when is_comparing_zero is truthy.
+        """
+        remove_if_zero = set(['linear', 'proportional'])
+        invalid = lambda mode: is_comparing_zero and mode in remove_if_zero
+        return tuple(mode for mode in self.modes if not invalid(mode) )
```
jolyonb: Minor linting at the end here. Can I suggest avoiding the double negative and doing an "allowed_if_zero" and "valid" instead?

mitxgraders/comparers/linear_comparer.py
ChristopherChudzicki: 👍

mitxgraders/comparers/linear_comparer.py
ChristopherChudzicki: 👍 new version is easier to read, I think

ChristopherChudzicki:
@jolyonb Just added some tests for MatrixGrader. Good thing, too. Found some bugs for it.

tests/comparers/test_linear_comparer.py
```diff
@@ -0,0 +1,109 @@
...
+        sample_from={
+            'A': RealMatrices(),
+            'B': RealMatrices()
+        },
+        max_array_dim=2
+    )
+
+    assert grader(None, 'x*A*B^2')['grade_decimal'] == 1.0
+    assert grader(None, '2*x*A*B^2')['grade_decimal'] == 0.6
+    assert grader(None, 'x*A*B^2 + 5*[[1, 1], [1, 1]]')['grade_decimal'] == 0.4
```
ChristopherChudzicki: @jolyonb This behavior bugs me... should be 5*Identity matrix that passes, but everything else with matrices works fine. (And vectors work fine).

I'm sure this is fixable, but I think that #200 is more important. I figure that in the changelog we mention that this quirk of LinearComparer + MatrixGrader might change.

jolyonb: Indeed, this is a bit of a weird one, but I suspect that even being off by an identity multiple is a weird situation for a matrixgrader.

mitxgraders/comparers/linear_comparer.py
```diff
@@ -154,29 +156,30 @@ def check_comparing_zero(comparer_params_evals, student_evals, tolerance):
...
         student_evals_norm = np.linalg.norm(student_evals)
 
         # Validate student input shape...only needed for MatrixGrader
         if hasattr(utils, 'validate_shape'):
             # in numpy, scalars have empty tuples as their shapes
-            scalar_expected = isinstance(comparer_params_evals[0], Number)
-            shape = tuple() if scalar_expected else comparer_params_evals[0].shape
+            expected_0 = comparer_params_evals[0][0]
+            scalar_expected = isinstance(expected_0, Number)
+            shape = tuple() if scalar_expected else expected_0.shape
```
jolyonb: This isn't a bug that has a corresponding issue in equality_comparer, is it?

jolyon:
Please could you take jolyon.wright@gmail.com off your mailing list!

I think my email address was added in error 

I have not subscribed!

Much appreciated,

Jolyon *w* :)

mitxgraders/comparers/linear_comparer.py
ChristopherChudzicki: No. Despite the same argument name, correlated comparers have an extra level of nesting.

In traditional comparers, `comparer_params_evals` is `[param0_eval, param1_eval,...]`.

In correlated comparers, `comparer_params_evals` is `[[param0_eval0, param1_eval0,...], [param_1_eval0, ...], ...]`

I should rename the argument somehow...

jolyonb:
My apologies Jolyon W! I honestly have no idea how that happened. I don't have control over which threads you've managed to get subscribed to; I can't find you at all from my end. I suspect there's an "unsubscribe" link at the bottom of any emails that you're getting though.

jolyonb:
Ah, I found it. You have @ jolyon while I have @ jolyonb (apparently you beat me to it!), and my colleague referenced you by mistake. I've fixed the reference; I think you can just unsubscribe yourself (or at least "mute the thread"). My apologies!

jolyonb:
I think you've addressed all my comments. Do you still have more you're working on here, or shall I do one final proofread?

ChristopherChudzicki:
One more commit. 

I just renamed most instances of `comparer_params_evals` to `comparer_params_eval`, unless the variable represents all evaluations of the comparer_params.

This brings comparers.py in line with FormulaGrader's usage of the terms.

Now it's ready.

jolyon:
Ha ha - ok

I do appreciate you following up on this :)

Best of luck with the project - I’m very flattered to be involved... in a a peripheral kind of a way !

Jolyon

Issue #197, opened 2019-06-23
Constants in DependentSampler


jolyonb:
DependentSampler sampling currently does not have access to the constants supplied in the problem. This mean, for instance, no access to i, pi, and e.

jolyonb:
Resolved by #198.

PR #198, opened 2019-06-23
give dependent sampler access to constants

ChristopherChudzicki:
addresses #197

### DIFF
--- a/mitxgraders/sampling.py
+++ b/mitxgraders/sampling.py
@@ -638,8 +638,11 @@ def gen_symbols_samples(symbols, samples, sample_from, functions, suffixes, cons
     sample_list = []
     for _ in range(samples):
         # Generate independent samples
-        sample_dict = {symbol: sample_from[symbol].gen_sample() for symbol in independent}
-        sample_dict.update(constants)
+        sample_dict = constants.copy()
+        sample_dict.update({
+            symbol: sample_from[symbol].gen_sample() for symbol in independent
+        })
+
         # Generate dependent samples, following chains as necessary
         unevaluated_dependents = {
             symbol: sample_from[symbol].config['depends'] for symbol in symbols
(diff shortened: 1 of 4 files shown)


mitxgraders/sampling.py
```diff
@@ -639,7 +639,7 @@ def gen_symbols_samples(symbols, samples, sample_from, functions, suffixes):
     for _ in range(samples):
         # Generate independent samples
         sample_dict = {symbol: sample_from[symbol].gen_sample() for symbol in independent}
-
+        sample_dict.update(constants)
```
jolyonb: I think this needs to come before the previous line (well, effectively). Otherwise, you can never override a constant.

jolyonb:
Thanks for doing this. Just the one comment.

mitxgraders/sampling.py
ChristopherChudzicki: good catch!

jolyonb:
Hmm. Sorry, I caught one more thing. `gen_symbols_samples` now returns dictionaries containing all the variables AND constants, right? So in `FormulaGrader`'s `raw_check` function, we don't need to initialize the variables with the constants any more; we can just set the variable list to the constants.

Also, there's a cute bug here that we need to fix: a dependent variable that has the same name as a constant can lead to problems. We need to scrub the list of constants of any variable names before we do the sampling.

mitxgraders/sampling.py
jolyonb: That's why we work in pairs!

jolyonb:
:+1:

ChristopherChudzicki:
I think this is all set now! Good catches...

> Also, there's a cute bug here that we need to fix: a dependent variable that has the same name as a constant can lead to problems. We need to scrub the list of constants of any variable names before we do the sampling.

Took me a little while to write an example.

jolyonb:
It's a cute little gotcha, isn't it? You'd have to be doing something fairly contrived to trigger it, but still!

PR #199, opened 2019-06-24
Test docs

ChristopherChudzicki:
This makes `pytest` run the doc/*.md files as doc tests, and fixes a few broken tests.

To facilitate this, I made `ObjectWithSchema.__repr__` pretty-format its `config` dictionary before printing. Primarily, the goal there was to make sure keys are displayed in alphabetical order (which `pprint` does for dictionaries).

### DIFF
--- a/docs/grading_math/sampling.md
+++ b/docs/grading_math/sampling.md
@@ -90,7 +90,9 @@ Sample real matrices of a specific shape and norm. (`RealMatrices` uses the Frob
 >>> # Sample 3 by 2 real matrices with norm between 5 and 10
 >>> sampler = RealMatrices(shape=[3, 2], norm=[5, 10])
 >>> # the default is shape=[2, 2] and norm=[1, 5]
->>> RealMatrices()
+>>> default_sampler = RealMatrices()
+>>> default_sampler
+RealMatrices({'norm': [1, 5], 'shape': (2, 2)})
 
 ```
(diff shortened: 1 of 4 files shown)


ChristopherChudzicki:
@jolyonb Came across this issue when playing with python 3 today. 

BTW: I suppose that there are cases where pretty-printing could be annoying by default, but in those cases I think our existing print method is already overwhelming. Anyway, I could move the pretty version into a separate method if you prefer.

ChristopherChudzicki:
@jolyonb As I mentioned, my primary goal in pretty-formatting the `ObjectWithSchema` config during `__repr__` was to alphabetize the keys. This is also going to help with py2/py3 compatibility since dictionary key order can be different in the two versions. (Dicts have a guaranteed print order in 3.7+, but not below.)

```python
>>> print({letter: letter for letter in 'abcd'})
{'a': 'a', 'b': 'b', 'c': 'c', 'd': 'd'} # in 3.7
{'a': 'a', 'c': 'c', 'b': 'b', 'd': 'd'} # in 2.7, on my machine anyway
```

jolyonb:
Just out of interest, you're using the master fork again?

ChristopherChudzicki:
> Just out of interest, you're using the master fork again?
Hmmm my remotes were messed up. Fixed them.

PR #205, opened 2019-06-27
Adding multiple features to StringGrader

### LINKED ISSUE #190: StringGrader enhancements
We could match to a regex expression and give a syntax error if it doesn't conform. Something that came up in my course... Also, a flag in StringGrader to strip all spaces from input.

jolyonb:
This PR adds a bunch of upgrades to `StringGrader`:

* Adds a minimum character count and a minimum word count when accepting all entries
* Adds new flags to determine how to treat whitespace
* Cleans up and expands the implementation of input cleaning
* Implements regex validation of inputs

While I was at it, I updated some docstrings in `baseclasses.py`.

This PR includes functionality and tests, but not documentation. I'm leaving that for later.

Resolves #190

### DIFF
--- a/mitxgraders/stringgrader.py
+++ b/mitxgraders/stringgrader.py
@@ -4,8 +4,11 @@ stringgrader.py
 Class for grading inputs that correspond to a text string
 * StringGrader
 """
-from voluptuous import Required
+import re
+from voluptuous import Required, Any
 from mitxgraders.baseclasses import ItemGrader
+from mitxgraders.helpers.validatorfuncs import NonNegative
+from mitxgraders.exceptions import InvalidInput, ConfigError
 
 # Set the objects to be imported from this grader
 __all__ = ["StringGrader"]
@@ -15,11 +18,54 @@ class StringGrader(ItemGrader):
     Grader based on exact comparison of strings
 
     Configuration options:
+        case_sensitive (bool): Whether to be case sensitive in comparing responses to
+            answers (default True)
+
+
+        clean_spaces (bool): Whether or not to convert multiple spaces into single spaces
+            before comparing (default True)
+
         strip (bool): Whether or not to strip leading and trailing whitespace
             from answers/student input before comparing (default True)
 
-        case_sensitive (bool): Whether to be case sensitive in comparing responses to
-            answers (default True)
+        strip_all (bool): Whether or not to remove all spaces from student
+            input before grading (default False)
+
+
+        accept_any (bool): Whether to accept any answer as correct (default False)
+
+        accept_nonempty (bool): Whether to accept any nonempty answer as correct.
+            Implemented as turning on accept_any and ensuring that min_length > 0.
+            (default False)
+
+
+        min_length (int): When using accept_any or accept_nonempty, sets the minimum
+            number of characters required to be entered in order to be graded
+            correct (default 0)
+
+        min_words (int): When using accept_any or accept_nonempty, sets the minimum
+            number of words required to be entered in order to be graded
+            correct (default 0)
+
+        explain_minimums (bool): If a response is unsatisfactory due to min_length or
+            min_words being insufficient, presents an explanatory message to the student
+            (default True)
+
+
+        validation_pattern (str or None): A regex pattern to validate the cleaned input
+            against. If the pattern is not satisfied, the setting in validation_err
+            is followed. Applies even when accept_any/accept_nonempty are True.
+            (default None)
+
+        validation_err (bool): Whether to raise an error if the validation_pattern
+            is not met (True) or to just grade as incorrect (False). If an error is
+            raised, invalid_msg is presented to students. When set to True, if debug is
+            also set to true, then a message is provided explaining that validation
+            failed. (default True)
+
+        invalid_msg (str): Error message presented to students if their input does
+            not satisfy the validation pattern
+            (default 'Your input is not in the expected format')
     """
 
     @property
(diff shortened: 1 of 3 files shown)


jolyonb:
Resolves #190.

mitxgraders/stringgrader.py
```diff
@@ -43,34 +128,75 @@ def check_response(self, answer, student_input, **kwargs):
...
+        if accept_any:
+            # Just check the minimums. No comparison to expect.
+            # Check for the minimum length
+            chars = len(student)
+            if chars < min_length:
+                if self.config['explain_minimums']:
+                    msg = 'Your response is too short ({chars}/{min_length} characters)'
+                    incorrect_response['msg'] = msg.format(chars=chars,
+                                                           min_length=min_length)
+                return incorrect_response
```
ChristopherChudzicki: Question: Would it make more sense to raise an error here, so as not to use up student's attempt? Or make it configurable: `explain_minimums=Any('msg', 'error', None)`?

I'm not too familiar with the use case for `accept_any` so I'm not sure what the preference would be.

mitxgraders/stringgrader.py
```diff
@@ -15,11 +18,54 @@ class StringGrader(ItemGrader):
     Grader based on exact comparison of strings
 
     Configuration options:
+        case_sensitive (bool): Whether to be case sensitive in comparing responses to
+            answers (default True)
+
```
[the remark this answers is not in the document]
jolyonb: Yeah, I was separating the configuration options into groups: case/spacing/accept any/minimums/validation. Perhaps I should put in headers to that extent?

mitxgraders/stringgrader.py
```diff
@@ -43,34 +128,75 @@ def check_response(self, answer, student_input, **kwargs):
             answer (str): The answer to compare to
             student_input (str): The student's input passed by edX
         """
-        expect = answer['expect']
```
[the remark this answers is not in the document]
jolyonb: Easy fix.

mitxgraders/stringgrader.py
jolyonb: The use case is typically "Write a sentence/paragraph on this topic", typically with unlimited submissions (because if you're accepting anything, does it really matter that they took a few submissions?). However, I do like your suggestion of raising an error rather than grading incorrect, or at least, having the option to do so. I'll implement it.

mitxgraders/stringgrader.py
```diff
@@ -43,34 +128,75 @@ def check_response(self, answer, student_input, **kwargs):
...
-        student = student_input
```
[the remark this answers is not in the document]
jolyonb: I'm going to push back on this one. The purpose of writing functions is typically so that you can reuse those functions elsewhere. In this case, all those functions would be called exactly once. Furthermore, most of those functions need to be able to return a grade, which is complicated by being in a subfunction. Finally, if you dropped the comments and didn't count all the lines associated with setting up feedback messages, this isn't really a long function at all. I'll do some refactoring to make it look shorter, and then see if you still feel that way.

mitxgraders/stringgrader.py
ChristopherChudzicki: The only downside I can think of is that errors don’t get logged for analytics (right?)

mitxgraders/stringgrader.py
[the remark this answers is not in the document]
ChristopherChudzicki: > The purpose of writing functions is typically so that you can reuse those functions elsewhere. 

Well, I agree that’s a great use of functions. But I would also suggest that a function with a readable, meaningful name that replaces a block of code is a good use, too, even if that function is only called once. For me, it makes the composite function’s structure much faster to understand. 

But, to each their own :) I’m OK reading long functions as long as you’re OK reading my single use functions 😛

mitxgraders/stringgrader.py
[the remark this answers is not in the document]
jolyonb: Modified appropriately.

mitxgraders/stringgrader.py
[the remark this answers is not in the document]
jolyonb: I think I've substantially improved the readability and flow of the method. Let me know what you think.

mitxgraders/stringgrader.py
jolyonb: So implemented.

jolyonb:
Glad you noticed the branch name ;-)

Should be ready for second review. Warning - the `check_response` method got a significant overhaul.

mitxgraders/stringgrader.py
jolyonb: I'm not sure the error text gets logged; I think the submission does though (and you can tell that it wasn't an attempt-counting submission as attempts doesn't increment).

mitxgraders/stringgrader.py
```diff
@@ -29,48 +75,123 @@ def schema_config(self):
...
+                    raise ConfigError(msg.format(answer['expect'], pattern))
+
+            # Check to see if the student input matches the validation pattern
+            if re.match(pattern, student) is None:
+                # Either raise an error or grade as incorrect
+                if self.config['validation_err']:
+                    raise InvalidInput(self.config['invalid_msg'])
+                else:
+                    if self.config['debug']:
+                        incorrect_response['msg'] = 'Validation failed'
```
ChristopherChudzicki: (Should have asked this on first review)

Why only add `config[‘invalid_msg’` when raising an error? Why not also show it when returning an incorrect response?

jolyonb: You might want to use a regex expression to say "this is the form that the right answer takes" instead of using an expect string. However, you raise a good point - I'll implement the same thing as for minimums.

jolyonb: And updated. You'll be really happy to see that I wrote a function!

PR #209, opened 2019-06-27
Update volutpuous

ChristopherChudzicki:
Updates voluptuous to version `0.11.5`.

Largely, this update is motivated by my desire to validate default values in dictionaries. In particular, some of our validators also transform the input. I was surprised today by:

```python
from mitxgraders import RealMatrices

x = RealMatrices()
y = RealMatrices({'norm': [1, 5], 'shape': (2, 2)})

print(x) # RealMatrices({'norm': [1, 5], 'shape': (2, 2)})
print(y) # RealMatrices({'norm': {'start': 1, 'stop': 5}, 'shape': (2, 2)})
```

### DIFF
--- a/docs/grading_math/sampling.md
+++ b/docs/grading_math/sampling.md
@@ -91,8 +91,8 @@ Sample real matrices of a specific shape and norm. (`RealMatrices` uses the Frob
 >>> sampler = RealMatrices(shape=[3, 2], norm=[5, 10])
 >>> # the default is shape=[2, 2] and norm=[1, 5]
 >>> default_sampler = RealMatrices()
->>> default_sampler
-RealMatrices({'norm': [1, 5], 'shape': (2, 2)})
+>>> default_sampler == RealMatrices(shape=[2, 2], norm=[1, 5])
+True
 
 ```
 
--- a/mitxgraders/formulagrader/formulagrader.py
+++ b/mitxgraders/formulagrader/formulagrader.py
@@ -947,8 +947,8 @@ class NumericalGrader(FormulaGrader):
             Required('user_functions', default={}): {Extra: is_callable},
             Required('tolerance', default='5%'): Any(PercentageString, NonNegative(Number)),
             Required('samples', default=1): 1,
-            Required('variables', default=[]): [],
-            Required('numbered_vars', default=[]): [],
+            Required('variables', default=[]): All(Length(max=0), []),
+            Required('numbered_vars', default=[]): All(Length(max=0), []),
             Required('sample_from', default={}): {},
             Required('failable_evals', default=0): 0
         })
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -55,6 +55,13 @@ class ObjectWithSchema(object):
         return "{classname}({config})".format(classname=self.__class__.__name__,
                                               config=pretty_config)
 
+    def __eq__(self, other):
+        """
+        Checks equality by checking class-equality and config equality.
+        """
+        return self.__class__ == other.__class__ and self.config == other.config
+
+
 class AbstractGrader(ObjectWithSchema):
     """
     Abstract grader class. All graders must build on this class.
--- a/mitxgraders/formulagrader/matrixgrader.py
+++ b/mitxgraders/formulagrader/matrixgrader.py
@@ -8,7 +8,7 @@ from collections import namedtuple
 from voluptuous import Required, Any
 from mitxgraders.exceptions import InputTypeError
 from mitxgraders.formulagrader.formulagrader import FormulaGrader
-from mitxgraders.helpers.validatorfuncs import NonNegative
+from mitxgraders.helpers.validatorfuncs import NonNegative, Nullable
 from mitxgraders.helpers.calc import MathArray, within_tolerance, identity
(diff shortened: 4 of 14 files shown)


mitxgraders/formulagrader/formulagrader.py
```diff
@@ -947,8 +947,8 @@ def schema_config(self):
             Required('user_functions', default={}): {Extra: is_callable},
             Required('tolerance', default='5%'): Any(PercentageString, NonNegative(Number)),
             Required('samples', default=1): 1,
-            Required('variables', default=[]): [],
-            Required('numbered_vars', default=[]): [],
+            Required('variables', default=[]): All(Length(max=0), []),
+            Required('numbered_vars', default=[]): All(Length(max=0), []),
```
jolyonb: Why do we need to add a maximum length of 0 to a list?

jolyonb: Ah, I'm guessing this is `NumericalGrader`...

jolyonb:
This all looks fine to me, but Travis is taking a long time (which I suspect means it's going to fail miserably).

> Largely, this update is motivated by my desire to validate default values in dictionaries. In particular, some of our validators also transform the input. I was surprised today by:
> 
> ```python
> from mitxgraders import RealMatrices
> 
> x = RealMatrices()
> y = RealMatrices({'norm': [1, 5], 'shape': (2, 2)})
> 
> print(x) # RealMatrices({'norm': [1, 5], 'shape': (2, 2)})
> print(y) # RealMatrices({'norm': {'start': 1, 'stop': 5}, 'shape': (2, 2)})
> ```

I can't quite parse your statement or your example. Can you elaborate?

jolyonb:
Ok, Travis just succeeded. I'm happy to merge, but if you can elaborate, that would be great ;-) (Also, I thought you'd be more excited about the inclusion of sets in voluptuous...)

docs/grading_math/sampling.md
```diff
@@ -91,8 +91,8 @@ Sample real matrices of a specific shape and norm. (`RealMatrices` uses the Frob
 >>> sampler = RealMatrices(shape=[3, 2], norm=[5, 10])
 >>> # the default is shape=[2, 2] and norm=[1, 5]
 >>> default_sampler = RealMatrices()
->>> default_sampler
-RealMatrices({'norm': [1, 5], 'shape': (2, 2)})
+>>> default_sampler == RealMatrices(shape=[2, 2], norm=[1, 5])
```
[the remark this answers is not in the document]
jolyonb: Ah, very good.

mitxgraders/formulagrader/formulagrader.py
```diff
@@ -947,8 +947,8 @@ def schema_config(self):
             Required('user_functions', default={}): {Extra: is_callable},
             Required('tolerance', default='5%'): Any(PercentageString, NonNegative(Number)),
             Required('samples', default=1): 1,
-            Required('variables', default=[]): [],
-            Required('numbered_vars', default=[]): [],
+            Required('variables', default=[]): All(Length(max=0), []),
```
ChristopherChudzicki: I believe that the usual empty-list testing, `Required('variables', default=[]): []`, is a bit broken in voluptuous `0.11.5`. I added `Length(max=0)` as a workaround, but this also gives a more informative error message IMO.

See https://github.com/alecthomas/voluptuous/issues/397

Despite this bug, I think upgrading is worth it.

ChristopherChudzicki:
> Can you elaborate?

@jolyonb The only difference between `x = RealMatrices()` and `y = RealMatrices({'norm': [1, 5], 'shape': (2, 2)})`  is that `y` had the default values explicitly passed. I was surprised that the configuration dictionaries were not equal. In `x` the default values were not being transformed, but in `y` since they are explicitly passed, the config values are validated+transformed. 

("Annoyed" is maybe a better word than "surprised", since I knew that some our validators transform and old voluptuous did not validate default values.)

ChristopherChudzicki:
(To be clearer, `x.config['norm']=[1, 5]` whereas `y.config['norm'] = { 'start': 1, 'stop': 5}`

jolyonb:
Ah, I see. So when you said "default values in dictionaries", you didn't mean dictionary entries, but the configuration dictionary as a whole?

ChristopherChudzicki:
>So when you said "default values in dictionaries", you didn't mean dictionary entries, but the configuration dictionary as a whole?

I think I meant the individual entries. For example, in the example below, the config dict as a whole is validated in the sense that all keys are validated, except the ones where default values are used by voluptuous to populate the dict.

But maybe we're trying to say the same thing 😛

```python
from voluptuous import Schema, All, Required

schema = Schema({
    Required('a', default='cat'): All(str, lambda s: s.upper()),
    Required('b', default=0): int
})

print(schema({'b': 5}))
# 0.10.5: {'a': 'cat', 'b': 5} 5 is verified to be an int, but 'ca't is NOT upper-cased.
# 0.11.5: {'a': 'CAT', 'b': 5} 5 is verified, AND the value 'CAT' is uppercase
```

jolyonb:
I think we are. I was confused because I thought you were thinking of something like `schema({'b': {'anotherdict': something}})` (ie, a dictionary for an individual element).

PR #200, opened 2019-06-24
First steps toward Python 2 / 3 compatibility

ChristopherChudzicki:

PR #210, opened 2019-06-29
Py23 unicode

ChristopherChudzicki:
Background: https://portingguide.readthedocs.io/en/latest/strings.html, merge after #200 

The goal of this PR is to make our library:
 - use unicode strings internally
 - allow Python 2 end-users to use `unicode` or `str`. This is achieved by the `text_string` voluptuous validator

Other notes:
 - Almost all `mitxgraders/` files use `from __future__ import unicode_literals`
     - The except is `__init__` files, since [`__all__` cannot contain unicode in Python 2](https://stackoverflow.com/questions/19913653/no-unicode-in-all-for-a-packages-init)  
 - No files in `tests/` use unicode_literals future import (since our end-users can't use future imports on edX)


jolyonb:
I'll get you to rebase this after #200 is merged. Then I'll review ;-)

jolyonb:
Was just going to ask you to close this and reopen a clean one ;-)

PR #211, opened 2019-06-29
Py23 unicode

ChristopherChudzicki:
Background: https://portingguide.readthedocs.io/en/latest/strings.html, merge after #200 

The goal of this PR is to make our library:
 - use unicode strings internally
 - allow Python 2 end-users to use `unicode` or `str`. This is achieved by the `text_string` voluptuous validator

Other notes:
 - Almost all `mitxgraders/` files use `from __future__ import unicode_literals`
     - The except is `__init__` files, since [`__all__` cannot contain unicode in Python 2](https://stackoverflow.com/questions/19913653/no-unicode-in-all-for-a-packages-init)  
 - No files in `tests/` use unicode_literals future import (since our end-users can't use future imports on edX)


jolyonb:
First thing I noticed is that test coverage is down, probably because of clauses that only trigger in python3. Something to keep in mind.

ChristopherChudzicki:
@jolyonb This is a re-do of #210 ... not sure why that one isn't showing the updated history after re-base. Anyway, this one shows a cleaner history.

BTW: Coverage has dropped slightly. I believe I can increase it, but it occurs to me that we may not be able to have 100% coverage in either Python 2 or Python 3 separately. I'll look into generating a merged coverage report later.

ChristopherChudzicki:
> First thing I noticed is that test coverage is down, probably because of clauses that only trigger in python3. Something to keep in mind.

You beat me by 30 seconds :(

mitxgraders/baseclasses.py
```diff
@@ -143,9 +145,9 @@ def __call__(self, expect, student_input):
         self.log("MITx Grading Library Version " + __version__)
         # Add the student inputs to the debug log
         if isinstance(student_input, list):
-            self.log("Student Responses:\n" + "\n".join(map(str, student_input)))
+            self.log("Student Responses:\n" + "\n".join(map(six.text_type, student_input)))
```
jolyonb: Ooh, ouch. That's an annoying change to have to deal with :-(

tests/plugins/test_integralgrader.py
```diff
@@ -516,7 +517,7 @@ def test_debug_message():
         "==============================================<br/>\n"
         "Integration Data for Sample Number 0<br/>\n"
         "==============================================<br/>\n"
-        "Variables: {{'infty': inf, 'e': 2.718281828459045, 'i': 1j, 'j': 1j, 's': 5.530375019455111, 'x': 5.530375019455111, 'pi': 3.141592653589793}}<br/>\n"
+        "Variables: {{{u}'infty': inf, {u}'e': 2.718281828459045, {u}'i': 1j, {u}'j': 1j, u's': 5.530375019455111, u'x': 5.530375019455111, {u}'pi': 3.141592653589793}}<br/>\n"
```
jolyonb: Wow, also unpleasant!

Is there a reason that s and x get `u` instead of `{u}`?

tests/plugins/test_integralgrader.py
```diff
@@ -540,7 +541,7 @@ def test_debug_message():
         "==============================================<br/>\n"
         "Integration Data for Sample Number 1<br/>\n"
         "==============================================<br/>\n"
-        "Variables: {{'infty': inf, 'e': 2.718281828459045, 'i': 1j, 'j': 1j, 's': 5.530375019455111, 'x': 5.530375019455111, 'pi': 3.141592653589793}}<br/>\n"
+        "Variables: {{{u}'infty': inf, {u}'e': 2.718281828459045, {u}'i': 1j, {u}'j': 1j, u's': 5.530375019455111, u'x': 5.530375019455111, {u}'pi': 3.141592653589793}}<br/>\n"
```
jolyonb: Similarly here.

tests/plugins/test_integralgrader.py
```diff
@@ -589,7 +590,7 @@ def test_debug_message_complex_integrand():
         "==============================================<br/>\n"
         "Integration Data for Sample Number 0<br/>\n"
         "==============================================<br/>\n"
-        "Variables: {{'infty': inf, 'e': 2.718281828459045, 'i': 1j, 'j': 1j, 's': 1.8831785881043805, 'x': 0.022981166359782736, 'pi': 3.141592653589793}}<br/>\n"
+        "Variables: {{{u}'infty': inf, {u}'e': 2.718281828459045, {u}'i': 1j, {u}'j': 1j, {u}'s': 1.8831785881043805, {u}'x': 0.022981166359782736, {u}'pi': 3.141592653589793}}<br/>\n"
```
jolyonb: But apparently not here. (This suggests to me that you can just use `u` on all of these, instead of needing the prefix variable?)

tests/formulagrader/test_formulagrader.py
```diff
@@ -466,7 +468,7 @@ def test_fg_config_expect():
 
     # If trying to use comparer, a detailed validation error is raised
     expect = ("to have 3 arguments, instead it has 2 for dictionary value @ "
-              "data\['answers'\]\[0\]\['expect'\]\['comparer'\]")
+              "data\['answers'\]\[0\]\[u?'expect'\]\['comparer'\]")
```
jolyonb: This I like :-)

tests/plugins/test_integralgrader.py
ChristopherChudzicki: The `u` instead of `{u}` was probably a mistake. They are equivalent in Python 2, and we're not really running the Python 3 tests yet. (A bunch of them fail for completely unrelated versions). 

So my guess is that this test would fail under Python 3. 

BTW: I am going to be gone for about 4-5 hours starting now-ish, so I'll just let you finish the review :)w

mitxgraders/sampling.py
```diff
@@ -668,8 +670,14 @@ def gen_symbols_samples(symbols, samples, sample_from, functions, suffixes, cons
         sample_list.append(sample_dict)
     return sample_list
 
+schema_user_functions_no_random = All(
```
jolyonb: Can we insert a comment explaining this?

mitxgraders/plugins/integralgrader.py
```diff
@@ -218,23 +220,18 @@ def schema_config(self):
...
             ),
             Required('tolerance', default='0.01%'): Any(PercentageString, NonNegative(Number)),
             Required('samples', default=1): Positive(int),  # default changed to 1
-            Required('variables', default=[]): All([str], all_unique),
+            Required('variables', default=[]): All([text_string], all_unique),
             Required('sample_from', default={}): dict,
             Required('failable_evals', default=0): NonNegative(int)
         })
 
-    schema_user_consts = All(
```
jolyonb: Were we just not using this?

mitxgraders/baseclasses.py
```diff
@@ -402,11 +459,6 @@ def check(self, answers, student_input, **kwargs):
                    "Expected at least one answer in answers")
             raise ConfigError(msg)
 
-        # Make sure the input is in the expected format
```
jolyonb: Is this now caught by `ensure_text_inputs`?

mitxgraders/baseclasses.py
```diff
@@ -205,6 +207,60 @@ def log_output(self):
         content = "\n".join(self.debuglog)
         return "<pre>{content}</pre>".format(content=content)
 
+    @classmethod
+    def ensure_text_inputs(cls, student_input, allow_lists=True, allow_single=True):
+        """
+        Ensures that student_input is a list of text strings.
```
jolyonb: Can't it just be a single string too (not wrapped in a list)?

mitxgraders/baseclasses.py
```diff
@@ -205,6 +207,60 @@ def log_output(self):
...
+
+        NOTE: Overriden by ItemGrader to ensure input is a single text string.
```
jolyonb: Add a note for `ListGrader` too?

mitxgraders/baseclasses.py
```diff
@@ -205,6 +207,60 @@ def log_output(self):
...
+            try:
+                return Schema([text_string])(student_input)
+            except MultipleInvalid as error:
+                pos = error.path[0]
+                list_errmsg = (
+                    "Expected a list of text strings for student_input, but "
+                    "item at position {pos} has {thetype}"
+                    .format(pos=pos, thetype=type(student_input[pos])))
+
+        if allow_lists and not isinstance(student_input, list):
```
jolyonb: Just cleaning this routine up a bit. I'd move this if statement below the `if allow_single:` below. See next comment.

mitxgraders/baseclasses.py
```diff
@@ -205,6 +207,60 @@ def log_output(self):
...
+                    "Expected string for student_input, received {}"
+                    .format(type(student_input)))
+
+        if allow_lists and allow_single:
+            msg = ("The student_input passed to a grader should be:\n"
+                   " - a text string for problems with a single input box\n"
+                   " - a list of text strings for problems with multiple input boxes\n"
+                   "Received student_input of {}").format(type(student_input))
+            raise ConfigError(msg)
+        elif allow_lists:
```
jolyonb: Put the text for this error message in here, instead of above. Same for the `allow_single` message.

mitxgraders/baseclasses.py
```diff
@@ -205,6 +207,60 @@ def log_output(self):
...
+            raise ConfigError(list_errmsg)
+        elif allow_single:
+            raise ConfigError(single_errmsg)
+        else:
+            raise ValueError('at least one of (allow_lists, allow_single) must be True.')
```
jolyonb: Picky minor lint - `At` instead of `at`.

mitxgraders/baseclasses.py
```diff
@@ -205,6 +207,60 @@ def log_output(self):
...
+
+        try:
```
jolyonb: It seems like you'll never get to here. Why is this code present?

tests/plugins/test_integralgrader.py
jolyonb: Ok, sounds good. Will let you update the missing entries then.

jolyonb:
I've gone through the whole PR now. Once you address the comments, I'll take a second pass. By and large, looking good though! (How badly are the python3 tests failing at this stage, out of curiosity?)

mitxgraders/baseclasses.py
```diff
@@ -205,6 +207,60 @@ def log_output(self):
...
+        """
+        list_errmsg = None
+        single_errmsg = None
+
+        if allow_lists and isinstance(student_input, list):
+            try:
+                return Schema([text_string])(student_input)
+            except MultipleInvalid as error:
+                pos = error.path[0]
+                list_errmsg = (
```
jolyonb: Any reason not to raise this immediately?

mitxgraders/baseclasses.py
jolyonb: Should be an `elif`, and once again, raise immediately.

mitxgraders/baseclasses.py
```diff
@@ -205,6 +207,60 @@ def log_output(self):
...
+        if allow_lists and not isinstance(student_input, list):
+            list_errmsg = (
+                "Expected student_input to be a list of text strings, but "
+                "received {}".format(type(student_input)))
+
+        if allow_single:
+            try:
+                return Schema(text_string)(student_input)
+            except MultipleInvalid as error:
+                single_errmsg = (
```
jolyonb: Again, raise immediately.

jolyonb: Or is the point not to raise this if both `allow_lists` and `allow_single` are True? Still, seems like too much spaghetti...

Perhaps something like the following?
```
        if allow_lists and isinstance(student_input, list):
            try:
                return Schema([text_string])(student_input)
            except MultipleInvalid as error:
                pass  # Will raise error below
        if allow_single:
            try:
                return Schema(text_string)(student_input)
            except MultipleInvalid as error:
                pass  # Will raise error below
```

This parses the valid inputs immediately, and you can select the appropriate error message to raise afterwards.

jolyonb:
Oh, I looked at the coverage report - there's a number of new lines in this PR that don't have coverage, but aren't associated with python3. A handful of new tests would catch them all. If you want to do them at a later stage, that's fine also.

tests/plugins/test_integralgrader.py
ChristopherChudzicki: fixed

tests/plugins/test_integralgrader.py
ChristopherChudzicki: changed to {u}

tests/plugins/test_integralgrader.py
ChristopherChudzicki: changed to {u}

tests/formulagrader/test_formulagrader.py
ChristopherChudzicki: Yeah, I think I put a note in the commit for the debug log, but I thought that replacing the debug log test with a regexp match would involve much too much escaping... {u} isn't too terrible.

mitxgraders/sampling.py
ChristopherChudzicki: added comment about NumericalGrader

mitxgraders/plugins/integralgrader.py
ChristopherChudzicki: That's correct, is was unused... the schema above used the imported `schema_user_functions`.

mitxgraders/baseclasses.py
ChristopherChudzicki: Yes. 

All graders now run `ensure_text_inputs` during `__call__`, which unicode or list of unicode. This is one of the few structural changes to the library in this PR. Without this change, I believe that each grader class would "need" to separately coerce input strings to unicode during `check`.

(I say "need" because it's really only an issue for local testing of library or local testing of course problem codes. edX gives unicode, so nothing on edX should ever really need that coercion. )

mitxgraders/baseclasses.py
ChristopherChudzicki: Yes, thanks.

I briefly considered having AbstractGrader demand a list of strings, not string or list of strings.

That feels inappropriate for AbstractGrader, but all single-input graders inherit from ItemGrader, right? No! I had forgotten that IntegralGrader can be multi-input or single-input.

mitxgraders/baseclasses.py
ChristopherChudzicki: 👍 
I'm also changing staticmethod to classmethod since `cls` is never used.

(I had erroneously thought staticmethods couldn't be called by super.)

mitxgraders/baseclasses.py
ChristopherChudzicki: 👍

mitxgraders/baseclasses.py
ChristopherChudzicki: It currently cannot be reached by using graders. It could be reached by a developer in the future who does something exceedingly silly:

```python
from mitxgraders.baseclasses import AbstractGrader

AbstractGrader.ensure_text_inputs('cat', allow_lists=False, allow_single=False)
# raises ValueError: At least one of (allow_lists, allow_single) must be True.
```

This is one of the lines that lacks coverage but is easy to test.

mitxgraders/baseclasses.py
ChristopherChudzicki: see below

mitxgraders/baseclasses.py
ChristopherChudzicki: see below

mitxgraders/baseclasses.py
ChristopherChudzicki: > Or is the point not to raise this if both allow_lists and allow_single are True? Still, seems like too much spaghetti...

Yes, that's why I avoided raising immediately.

I'm not sure that the structure you're thinking of will work without nesting the try/excepts fairly heavily, since the error variables are local:

```python
try:
    1/0
except ZeroDivisionError as error1:
    print(error1) # prints division by zero
    pass

try:
    import puppy
except ImportError as error2:
    print(error2) # pritns No module named 'puppy'
    pass

print(error1) # raises NameError: name 'error1' is not defined
print(error2)
```

Let me see if I can think about other ways to restructure this.

ChristopherChudzicki: Well, I think the new version is cleaner, but I'm sure you'll like it more. If you have other ideas, or if I misunderstood your bit about saving and raising errors, let me know.

ChristopherChudzicki:
> Oh, I looked at the coverage report - there's a number of new lines in this PR that don't have coverage, but aren't associated with python3. A handful of new tests would catch them all. If you want to do them at a later stage, that's fine also.

I'll write more tests tomorrow morning, then you can merge if approve changes.

ChristopherChudzicki:
Oh, re
> How badly are the python3 tests failing at this stage, out of curiosity?

Not too bad. Right now with Python 3:

```
44 failed, 257 passed, 1 error in 7.15 seconds
```

But 31 of the 44 failing tests are doctests—I'm pretty most of those 31 are failing because error-raising doctests need to be re-written as try/except since Python 3 prints errors slightly differently.

Without the doctests, only 13 tests are failing, and for several of those I know why already.

mitxgraders/baseclasses.py
```diff
@@ -207,39 +207,76 @@ def log_output(self):
...
         return "<pre>{content}</pre>".format(content=content)
 
-    @classmethod
-    def ensure_text_inputs(cls, student_input, allow_lists=True, allow_single=True):
+    @staticmethod
+    def ensure_list_of_text_inputs(student_input):
         """
-        Ensures that student_input is a list of text strings.
+        Delegated to by AbstractGrader.ensure_text_inputs. Validates whether
+        student_input is a list of text strings.
```
jolyonb: Again, is a list or is a text string or list of text strings?

mitxgraders/baseclasses.py
```diff
@@ -207,39 +207,76 @@ def log_output(self):
...
 
-        NOTE: Overriden by ItemGrader to ensure input is a single text string.
+        Returns:
+        A pair (result, errmsg) where either:
+            - result is a list of text strings and errmsg is None, or
```
jolyonb: A pair??? That... doesn't make sense. It returns either the result, or raises an error message if something went wrong...

mitxgraders/baseclasses.py
jolyonb: Oh, you split it into two methods.

mitxgraders/baseclasses.py
```diff
@@ -207,39 +207,76 @@ def log_output(self):
...
-                return Schema(text_string)(student_input)
-            except MultipleInvalid as error:
-                single_errmsg = (
-                    "Expected string for student_input, received {}"
-                    .format(type(student_input)))
+            result, single_errmsg = AbstractGrader.ensure_single_text_inputs(student_input)
+            if single_errmsg is None:
+                return result
 
+        # An error happened. Decide the correct case and raise.
```
jolyonb: I think this actually complicated things. I'll send you what I had in mind...

mitxgraders/baseclasses.py
jolyonb: But everything should be caught by the `else` statement above...

jolyonb:
Here's my proposed function:
```python
    @staticmethod
    def ensure_text_inputs(student_input, allow_lists=True, allow_single=True):
        """
        Ensures that student_input is a text string or a list of text strings,
        depending on arguments. Called by ItemGrader and ListGrader with 
        appropriate arguments. Defaults are set to be friendly to user-defined
        grading classes.
        """
        # Try to perform validation
        try:
            if allow_lists and isinstance(student_input, list):
                return Schema([text_string])(student_input)
            elif allow_single and not isinstance(student_input, list):
                return Schema(text_string)(student_input)
        except MultipleInvalid as error:
            if allow_lists:
                pos = error.path[0]

        # The given student_input is invalid, so raise the appropriate error message
        if allow_lists and allow_single:
            msg = ("The student_input passed to a grader should be:\n"
                   " - a text string for problems with a single input box\n"
                   " - a list of text strings for problems with multiple input boxes\n"
                   "Received student_input of {}"
                   ).format(type(student_input))
        elif allow_lists and not isinstance(student_input, list):
            msg = ("Expected student_input to be a list of text strings, but "
                   "received {}"
                   ).format(type(student_input))
        elif allow_lists:
            msg = ("Expected a list of text strings for student_input, but "
                   "item at position {pos} has {thetype}"
                   ).format(pos=pos, thetype=type(student_input[pos])))
        elif allow_single:
            msg = ("Expected string for student_input, received {}"
                   ).format(type(student_input))
        else:
            raise ValueError('At least one of (allow_lists, allow_single) must be True.')

        raise ConfigError(msg)
```
I think this is really clean. What do you think?

ChristopherChudzicki:
Yes, that's better 🎉

ChristopherChudzicki:
@jolyonb ready for re-review

By the way: Do you think edX would be open to upgrading their `six` from `1.11.0` to `1.12.0`? If they did, we could use `six`'s version of `ensure_text_inputs`. (I just copied it from the 1.12.0 repo).

jolyonb:
I think you meant a different six function that you copied? Regardless, we need to keep six at 1.11.0, as that's the version that old editions of edX will be running. Going to go and do my final review now.

jolyonb:
:+1:

Issue #212, opened 2019-06-30
Coveralls report seems to be stuck in July 2018?


jolyonb:
Go to front page, click on "coverage" badge - it reports things from July 2018, and nothing recent. Should look into this...

jolyonb:
Seems to be fixed since I logged into coveralls... Not sure what was happening.

PR #214, opened 2019-07-01
Updating StringGrader documentation

jolyonb:
What it says on the tin.

Also updates `python_lib.zip`, and removes an old reference in `.coveragerc`.


jolyonb:
Pushed to mkdocs for testing purposes.

Alas, note that mkdocs doesn't do pycon quite so nicely (it still lets you select the `>>>` and `...`)

jolyonb:
Also note that we now have two Travis checks running. One is on this branch, the other is against a merge into master. I think this might fix #212.

jolyonb:
@ChristopherChudzicki Ready for review. I'll update the course examples in a separate PR.

docs/index.md
```diff
@@ -148,3 +148,10 @@ Any `.py` file stored in the `plugins` folder will be automatically loaded. All
 You can define custom grading classes in your plugin. To learn how this works, we recommend copying the code from `stringgrader.py`, renaming the class, and building a simple plugin based on `StringGrader`.
 
 We are happy to include user-contributed plugins in the repository for this library. If you have built a plugin that you would like to see combined into this library, please contact the authors through [github](https://github.com/mitodl/mitx-grading-library). We are also willing to consider incorporating good plugins into the library itself.
+
+
+## Testing docs
+
+We're trying to make all of our examples test against the library. To invoke tests of the documentation, you can run the following command:
+
+`python -m pytest --no-cov --disable-warnings docs/*`
```
jolyonb: I wanted to put this somewhere so that we (or at least, I) could find it easily in the future!

ChristopherChudzicki: Good idea. Maybe a better spot would be in https://github.com/mitodl/mitx-grading-library#local-installation as a parenthetical to the 3rd list item?

docs/index.md
```diff
@@ -148,3 +148,10 @@ Any `.py` file stored in the `plugins` folder will be automatically loaded. All
 You can define custom grading classes in your plugin. To learn how this works, we recommend copying the code from `stringgrader.py`, renaming the class, and building a simple plugin based on `StringGrader`.
 
 We are happy to include user-contributed plugins in the repository for this library. If you have built a plugin that you would like to see combined into this library, please contact the authors through [github](https://github.com/mitodl/mitx-grading-library). We are also willing to consider incorporating good plugins into the library itself.
+
```
ChristopherChudzicki: Just noticed this line above (unedited in this pr):

`You can define custom grading classes in your plugin. To learn how this works, we recommend copying the code from `stringgrader.py`, renaming the class, and building a simple plugin based on `StringGrader`.`

`StringGrader` started (2 years ago?) as a toy example, I guess it's still simple enough to be a good entry point, though.

docs/string_grader.md
```diff
@@ -1,66 +1,236 @@
...
 ```
 
 This will accept `Cat`, `cat` and `CAT`. By default, `case_sensitive=True`.
 
 
-## Stripping Input
+## Accepting anything (within reason!)
 
-Leading or trailing spaces in an answer rarely change the meaning of the answer. Hence, by default, we strip all leading and trailing spaces from the student input and author-specified answers before comparison. If you want to keep those spaces around for the comparison, you need to disable strip.
+Sometimes you may just want to accept anything that a student provides (possibly subject to conditions). To do this, to set the `accept_any` flag, which will literally accept anything that is entered into the textbox.
```
ChristopherChudzicki: Is it worth adding something about the motivation? "...perhaps for a free-response question". Feel free to reject.

docs/string_grader.md
```diff
@@ -1,66 +1,236 @@
...
 
-## Accepting Anything
+```
 
-Sometimes you may just want to accept anything that a student provides. There are two options that allow you to do this. The first is `accept_any`, which will literally accept anything that is entered into the textbox. The second is `accept_nonempty`, which requires the textbox to contain a non-empty submission (blank spaces are stripped). `accept_any` has priority over `accept_nonempty`.
+Note that punctuation doesn't break a word for the purpose of word counting, so `isn't word-counting fun?` will only count as three words. If `accept_nonempty` and `min_length` are both used, the more stringent requirement is the one that is used.
+
+When a student's answer is rejected because it doesn't meet the minimum requirements, there are three types of feedback that you can provide, controlled by the `explain_minimums` flag:
+
+* The student receives an error message describing how many words/characters they have, compared to how many are required. This does not consume an attempt (`explain_minimums='err'`, default)
```
ChristopherChudzicki: I just noticed that the `accept_nonempty=True` example above did not raise an error. Should it? Conceptually it falls int he case of "When a student's answer is rejected because it doesn't meet the minimum requirements..."

Maybe `accept_nonempty=True` should internally just modify the configuration during `__init__` so that `StringGrader(accept_nonempty=True)` is exactly equivalent to `StringGrader(accept_any=True, min_length=1)`?

docs/string_grader.md
```diff
@@ -1,66 +1,236 @@
...
+...     validation_pattern=r'\([0-9]\)\([0-9]\)',
+...     explain_validation='err',
+...     strip_all=True   # Removes all spaces from the input
+... )
+>>> grader(None, '(1)(2)') == {'grade_decimal': 1, 'msg': '', 'ok': True}
+True
+>>> grader(None, '(2)(1)') == {'grade_decimal': 0, 'msg': '', 'ok': False}
+True
+>>> grader(None, '(a)(2)')
+Traceback (most recent call last):
```
ChristopherChudzicki: Sadly this doesn't work in Python 3. You can change it if you want, or you can consider that my problem 😛 

Python 3 prints traceback ends with `mitxgraders.exceptions.InvalidInput` instead of `InvalidInput`.

See http://python3porting.com/problems.html#handling-expected-exceptions

(Possibly this could be resolved by modifing sys. excepthook, but this seems hard and I can't find anyone else who's done it online.)

docs/string_grader.md
```diff
@@ -1,66 +1,239 @@
...
+...     answers='(1)(2)',
+...     validation_pattern=r'\([0-9]\)\([0-9]\)',
+...     explain_validation='err',
+...     strip_all=True   # Removes all spaces from the input
+... )
+>>> grader(None, '(1)(2)') == {'grade_decimal': 1, 'msg': '', 'ok': True}
+True
+>>> grader(None, '(2)(1)') == {'grade_decimal': 0, 'msg': '', 'ok': False}
+True
+>>> from mitxgraders.exceptions import InvalidInput
```
ChristopherChudzicki: I moved to `*` imports for the matrix docs so as not to have to import the errors, since importing the error isn't really doc-related.

docs/string_grader.md
ChristopherChudzicki: Thanks :)

docs/string_grader.md
jolyonb: Ok, will do.

docs/string_grader.md
jolyonb: Should be fixed now, I think...

docs/string_grader.md
jolyonb: So added.

docs/index.md
jolyonb: So done.

docs/index.md
jolyonb: Indeed, I remembered that we had that there... It's still a much simpler grader than literally anything else that we have, so I'm happy to keep that comment there.

docs/string_grader.md
jolyonb: I had `explain_minimums=None` turned on, which explains your first statement. As for the second, `accept_nonempty` internally sets `accept_any=True` and `min_length = max(1, min_length)`. I would get rid of `accept_nonempty` completely, were it not for backwards compatibility...

docs/string_grader.md
jolyonb: I had to add `InvalidInput` to `__init__.py` for this to work. Worth it though, I think.

docs/string_grader.md
ChristopherChudzicki: Ah, I should hav noticed that `explain_minimums=None`. My bad.

jolyonb: Note that throughout the docs, I have explanations set to all of `None`, `err` and `msg`, so each has an example of its behavior. Let me know if you think I should change the above though.

jolyonb:
I think this is ready for re-review.

PR #216, opened 2019-07-01
Adding floor and ceiling functions

### LINKED ISSUE #191: Floor and Ceiling functions
So, I had a request for these... We do need to document the behavior of them for negative numbers carefully though. Might as well add in a `round` function while we're at it.

jolyonb:
What it says on the tin.

Resolves #191.


jolyonb:
@ChristopherChudzicki Ready! This should be really quick.

PR #217, opened 2019-07-01
Updating SingleListGrader to detect missing entries

### LINKED ISSUE #167: SingleListGrader and empty entries
If a list is entered as `1, 2, , 3`, this should probably generate an error due to a missing entry. Similarly for `1, 2, 3,`.

jolyonb:
Adds an option to SingleListGrader to detect and present an error if there are empty entries in a list, turned on by default.

Includes tests and documentation (I started converting the examples to doctests).

Resolves #167.

### DIFF
--- a/docs/grading_lists/single_list_grader.md
+++ b/docs/grading_lists/single_list_grader.md
@@ -2,18 +2,26 @@
 
 If you want a response to be a delimiter-separated list of items, you can use a special ItemGrader called `SingleListGrader` to perform the grading. You need to specify a subgrader (which must be an ItemGrader, and could even be another SingleListGrader) to evaluate each individual item. The basic usage is as follows.
 
-```python
-grader = SingleListGrader(
-    answers=['cat', 'dog'],
-    subgrader=StringGrader()
-)
+```pycon
+>>> from mitxgraders import *
+>>> grader = SingleListGrader(
+...     answers=['cat', 'dog'],
+...     subgrader=StringGrader()
+... )
+>>> grader(None, 'cat, dog') == {'grade_decimal': 1.0, 'msg': '', 'ok': True}
+True
+>>> grader(None, 'dog, cat') == {'grade_decimal': 1.0, 'msg': '', 'ok': True}
+True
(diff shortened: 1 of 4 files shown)


mitxgraders/listgrader.py
```diff
@@ -692,6 +696,20 @@ def check_response(self, answer, student_input, **kwargs):
         answers = answer  # Rename from the ItemGrader name
         student_list = student_input.split(self.config['delimiter'])
 
+        # Check for empty entries in the list
+        if self.config['missing_error']:
+            bad_items = []
```
ChristopherChudzicki: `bad_items = [idx+1 for (idx, item) in enumerate(student_list) if item.strip() == '']` if you want.

docs/grading_lists/single_list_grader.md
```diff
@@ -2,18 +2,26 @@
...
+True
+
 ```
 
 To receive full points for this problem, a student would enter `cat, dog` or `dog, cat` into the input box. Entering `cat, octopus` or just `cat` will receive half points. The `answers` key is a python list of the individual answers you wish to use.
 
 You can use a tuple of lists to specify multiple lists of answers, just like normal ItemGraders.
 
-```python
+```pycon
```
ChristopherChudzicki: I see change python -> pycon, but you still need the prompts `>>>` and `...` in order to make this run as a doctest. Also, I suspect the `pycon` syntax highlighting will look wrong without the prompts, but I am not sure.

(Oops, I guess I never submitted this comment in earlier review)

jolyonb:
I think the problem is worst in SingleListGrader, because it's too easy to have an errant comma. In a ListGrader scenario, it's possible to leave a box empty, but I'm much less worried about that (sometimes it's by design).

mitxgraders/listgrader.py
jolyonb: Yeah, that is neater.

docs/grading_lists/single_list_grader.md
jolyonb: I think I was ambitious and thought I'd update the entire doc, but never quite got around to it... Well, maybe I'll just do so now.

jolyonb:
Implemented your suggestion and updated the documentation properly. Ready for a final review.

PR #220, opened 2019-07-02
Adding the ApplyFunctionComparer comparer

### LINKED ISSUE #192: New comparer: apply_function_comparer
I've run into a need to apply a function to both input and answer before comparing for equality. In my case, the functions are `re` and `im`, but I can imagine other uses too (eg, `mod`). This should be really straightforward to implement.

jolyonb:
Creates a comparer that applies a python function to the expected result and student input before performing the comparison. I thought about doing this based on string references to functions the grader already knows about, before realizing that comparers don't have access to the function listing etc.

In order to implement this, I created a new base class `Comparer`, of which `CorrelatedComparer` is a subclass. I also made an `EqualityComparer` class from which `equality_comparer` derives.

Question: Do you think that I should just merge `ApplyFunctionComparer` into `EqualityComparer`? They're almost copies of each other.

Resolves #192.

### DIFF
--- a/mitxgraders/comparers/__init__.py
+++ b/mitxgraders/comparers/__init__.py
@@ -1,24 +1,29 @@
 from __future__ import print_function, division, absolute_import
 
 from mitxgraders.comparers.comparers import (
+    EqualityComparer,
     equality_comparer,
+    ApplyFunctionComparer,
     congruence_comparer,
     eigenvector_comparer,
     between_comparer,
     vector_span_comparer,
     vector_phase_comparer
 )
-from mitxgraders.comparers.baseclasses import CorrelatedComparer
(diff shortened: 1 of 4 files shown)


ChristopherChudzicki:
> Question: Do you think that I should just merge ApplyFunctionComparer into EqualityComparer? They're almost copies of each other.

Yes, how about just a configuration key of `transform` or `transformer` or something on `EqualityComparer`, rather than a separate class. The key could either default to `None` (would need a logic tweak) or to the identity function.

jolyonb:
Ok, updated. Feel free to take a look now.

jolyonb:
Travis finished! =D

PR #223, opened 2019-07-03
Py23 all passing

ChristopherChudzicki:
(Hopefully) final code updates for Python 2 / 3 compatibility. All tests are passing in Python 2 and Python 3.

Re coverage: visual inspection shows no line missing in both coverage reports (Py2, Py3)

Going to look into automating the simultaneous testing / combined coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,23 @@
 # Travis continuous integration configuration file
 matrix:
   include:
-  - dist: xenial
   - language: python
     python:
       - "2.7"
-      - "3.7"
     install:
       - pip install -r requirements.txt
+      - pip install python-coveralls
+    script:
+      - pytest --cov=mitxgraders --cov-report=term-missing
+    after_success:
+      - coveralls
+
+  - language: python
+    python:
+      - "3.7"
+    dist: xenial
+    install:
+      - pip install -r requirements3.txt
       - pip install python-coveralls
     script:
       - pytest --cov=mitxgraders --cov-report=term-missing
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 # Travis continuous integration configuration file
 matrix:
   include:
+  - dist: xenial
   - language: python
     python:
       - "2.7"
+      - "3.7"
     install:
       - pip install -r requirements.txt
       - pip install python-coveralls
--- a/mitxgraders/comparers/linear_comparer.py
+++ b/mitxgraders/comparers/linear_comparer.py
@@ -18,12 +18,12 @@ def get_linear_fit_error(x, y):
     Zero error in a linear relationship:
     >>> x = np.array([2, 5, 8])
     >>> result = get_linear_fit_error(x, 2*x + 1)
-    >>> round(result, 8)
+    >>> round(result, 6)
     0.0
     """
     A = np.vstack([x, np.ones(len(x))]).T
     _, residuals, _, _ = np.linalg.lstsq(A, y, rcond=-1)
-    return np.sqrt(residuals)
+    return np.sqrt(np.linalg.norm(residuals))
 
 def get_proportional_fit_error(x, y):
     """
@@ -38,17 +38,17 @@ def get_proportional_fit_error(x, y):
     Reveals error if relationship is not proportional:
     >>> x = np.array([2, 5, 8])
     >>> result = get_proportional_fit_error(x, 2*x + 1)
-    >>> round(result, 8)
-    0.76200076
+    >>> result                                 # doctest: +ELLIPSIS
+    0.76200...
 
     Zero error in a proportional relationship:
     >>> result = get_proportional_fit_error(x, 2*x)
-    >>> round(result, 8)
+    >>> round(result, 6)
     0.0
     """
     A = np.vstack(x)
     _, residuals, _, _ = np.linalg.lstsq(A, y, rcond=-1)
-    return np.sqrt(residuals)
+    return np.sqrt(np.linalg.norm(residuals))
 
 def get_offset_fit_error(x, y):
     """
@@ -63,12 +63,12 @@ def get_offset_fit_error(x, y):
     Reveals error if relationship is not constant-offset:
     >>> x = np.array([2, 5, 8])
     >>> result = get_offset_fit_error(x, 2*x + 1)
-    >>> round(result, 8)
-    4.24264069
+    >>> result                                              # doctest: +ELLIPSIS
+    4.242640...
 
     Zero error in a constant-offset relationship:
     >>> result = get_offset_fit_error(x, x + 5)
-    >>> round(result, 8)
+    >>> round(result, 6)
     0.0
     """
     mean = np.mean(y - x)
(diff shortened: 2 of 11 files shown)


jolyonb:
I strongly suggest making a new PR to sort out coverage and Travis. I posted a link on your issue about build stages; I believe the idea is to set up a build stage for tests, and a second build stage to combine coverage reports before running coveralls.

mitxgraders/comparers/linear_comparer.py
```diff
@@ -18,12 +18,12 @@ def get_linear_fit_error(x, y):
...
     >>> x = np.array([2, 5, 8])
     >>> result = get_linear_fit_error(x, 2*x + 1)
-    >>> round(result, 8)
+    >>> round(result, 6)
     0.0
     """
     A = np.vstack([x, np.ones(len(x))]).T
     _, residuals, _, _ = np.linalg.lstsq(A, y, rcond=-1)
-    return np.sqrt(residuals)
+    return np.sqrt(np.linalg.norm(residuals))
```
jolyonb: Careful here. Are you sure you want the square root of the norm? I think that's already square rooted?

I'm guessing numpy changed what it returned for residuals?

ChristopherChudzicki: Good catch.

> I'm guessing numpy changed what it returned for residuals?

No. Numpy always returns an array for the residuals, but in early versions `round(np.array([ 1.23456])` would return a float, whereas in later versions it raises an error.

This caused problems for a doctest.

Besides testing issue, it bugs me that two of the `get_error...` functions return numbers, while the other two return numpy arrays (with 1 element).

ChristopherChudzicki:
> I strongly suggest making a new PR to sort out coverage and Travis. I posted a link on your issue about build stages; I believe the idea is to set up a build stage for tests, and a second build stage to combine coverage reports before running coveralls.

Yes, I'll definitely to a separate PR for this, although it seems like making travis run multiple versions of Python is easy so I just added that in my last commit. Maybe should have waited...but I'm eager to see it pass 😄

mitxgraders/comparers/linear_comparer.py
```diff
@@ -18,12 +18,12 @@ def get_linear_fit_error(x, y):
...
     >>> x = np.array([2, 5, 8])
     >>> result = get_linear_fit_error(x, 2*x + 1)
-    >>> round(result, 8)
+    >>> round(result, 6)
     0.0
     """
     A = np.vstack([x, np.ones(len(x))]).T
     _, residuals, _, _ = np.linalg.lstsq(A, y, rcond=-1)
-    return np.sqrt(residuals)
+    return np.sqrt(np.asscalar(residuals))
```
ChristopherChudzicki: @jolyonb Actually, I'm changing this to `np.asscalar`... residuals is always a 1-item array since we flatten the student input. 

BTW: `residuals` is actually the squared residuals, but I called it residuals b/c that's what numpy does https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.lstsq.html

jolyonb: Ok, `asscalar` makes more sense to me!

jolyonb:
Waaait, why is travis doing a ruby test run?

ChristopherChudzicki:
> Waaait, why is travis doing a ruby test run?

I was just wondering that. BTW, the `dist: xenial` bit in `.travis.yml` is in order to run Python 3.7: https://docs.travis-ci.com/user/languages/python/#python-37-and-higher

Also, Travis is slow: https://www.traviscistatus.com/

jolyonb:
Travis and Ruby: http://katyhuff.github.io/python-testing/extras/discuss.html

Might be a typo in the `.travis.yml` file?

.travis.yml
```diff
@@ -1,9 +1,11 @@
 # Travis continuous integration configuration file
 matrix:
   include:
+  - dist: xenial
```
jolyonb: So, the fact that this starts with a dash means that it's an included run in the matrix all by itself. What you want to do is put `dist: xenial` right at the top of the file, methinks.

.travis.yml
```diff
@@ -1,9 +1,11 @@
...
   - language: python
     python:
       - "2.7"
+      - "3.7"
```
jolyonb: This isn't how to include a new build in a build matrix, sorry. You need to copy the python block. (Build matrices have different syntax to normal travis files.)

mitxgraders/helpers/validatorfuncs.py
```diff
@@ -104,10 +104,10 @@ def all_unique(iterable):
     items should be unique, but have unexpected duplicates: ['a', 0]
     """
     seen = set()
-    duplicates = set()
+    duplicates = []
     for item in iterable:
         if item in seen:
-            duplicates.add(item)
+            duplicates.append(item)
```
jolyonb: I think this doesn't work if you pass in ['a', 0, 0, 0], as 0 gets included in duplicates twice. That's why we were using a `set`.

jolyonb: Probably means we need to add that case as a new test.

ChristopherChudzicki:
> Minor comment on one change (list -> set). Otherwise, once the travis config is fixed, this looks good to go.

Sounds good. I'm going to wait for Travis to finish #224  the other job before changing the config file again though.

mitxgraders/helpers/validatorfuncs.py
```diff
@@ -96,20 +97,19 @@ def all_unique(iterable):
...
+    >>> iterable = ['a', 0, '1', 'a', '5', 0, 0, '0']
     >>> try:
     ...     all_unique(iterable)
     ... except Invalid as error:
     ...     print(error)
     items should be unique, but have unexpected duplicates: ['a', 0]
     """
-    seen = set()
-    duplicates = set()
+    counts = defaultdict(int)
```
ChristopherChudzicki: The `set` version of `duplicates` wasn't good enough, because it is unordered and generally cannot be sorted by `sorted` (e.g., in the doctest above where strings and integers are incomparable for sorting).

So this version keeps `duplicates` sorted by order in the original iterable object.

mitxgraders/helpers/validatorfuncs.py
ChristopherChudzicki: The set version of duplicates wasn't good enough, because it is unordered and generally cannot be sorted by sorted (e.g., in the doctest above where strings and integers are incomparable for sorting).

So this version keeps duplicates sorted by order in the original iterable object.

Updated the test to have two copies of `0`

ChristopherChudzicki:
A test is failing on travis, but not locally. Travis is using Python `3.7.1`, whereas I'm on `3.7.3`.

If that test really depends on a patch version, I might remove it. That test is a pain in the butt.

jolyonb:
Ok, I'm looking at the test.

jolyonb:
Tell you what, why don't you just remove all the mocked stuff from that test (including the `assert_called_with` statement), and just pass in the custom comparer directly? That should fix the issue. The only thing that mock is getting us is that assert statement, and it's not like it's critical.

ChristopherChudzicki:
Works for me locally on 3.7.1. Looking more

ChristopherChudzicki:
> Tell you what, why don't you just remove all the mocked stuff from that test (including the assert_called_with statement), and just pass in the custom comparer directly? That should fix the issue. The only thing that mock is getting us is that assert statement, and it's not like it's critical.

Agreed

ChristopherChudzicki:
The branch name is finally true on Travis 🎉 😄 🎉

jolyonb:
Yay! The coverage decrease is due, I think, to the python2 running and then the python3, with both reporting to coveralls. Let's fix that in a separate PR.

PR #224, opened 2019-07-03
Py23 travis [WIP]

ChristopherChudzicki:


.travis.yml
```diff
@@ -12,6 +12,18 @@ matrix:
     after_success:
       - coveralls
 
+  - language: python
```
jolyonb: This looks like the correct block to me. Can you go and fix the other PR?

ChristopherChudzicki: Sure, or do you want to review this one and close the other? History is slightly cleaner. 

But if you have a bunch of unpublished comments I can go fix the other. 

(If you want to review this branch, I won't mess with travis anymore till after lunch)

jolyonb: I just finished reviewing the other branch...

ChristopherChudzicki: No problem :) I'm going to wait for this to finish before changing travis on the other branch.

jolyonb:
Well, it failed. Looks like you want requirements3.txt, no? Also, you may need to use pip3 instead of pip. But try these things in the other branch ;-)

jolyonb:
Actually, pip seems fine - it just looks like the version of numpy we use in edX isn't compatible with python3 ;-)

ChristopherChudzicki:
> Well, it failed. Looks like you want requirements3.txt, no? 

Ugh, you're right. Well, at least it ran Python 3.7. Fixing requirements3 on the other branch.

PR #226, opened 2019-07-03
try combining coverage with coveralls

ChristopherChudzicki:
Trying: https://docs.coveralls.io/parallel-build-webhook not from fork

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
       - "2.7"
     install:
       - pip install -r requirements.txt
-      - pip install --user codecov
+      - pip install codecov
     script:
       - pytest --cov=mitxgraders --cov-report=term-missing
     after_success:
@@ -18,7 +18,7 @@ matrix:
     dist: xenial
     install:
       - pip install -r requirements3.txt
-      - pip install --user codecov
+      - pip install codecov
     script:
       - pytest --cov=mitxgraders --cov-report=term-missing
     after_success:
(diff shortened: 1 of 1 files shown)


jolyonb:
Lol, you killed coveralls...

.travis.yml
```diff
@@ -7,8 +7,11 @@ matrix:
     install:
       - pip install -r requirements.txt
       - pip install python-coveralls
+    env: COVERALLS_PARALLEL=true
     script:
       - pytest --cov=mitxgraders --cov-report=term-missing
+    notifications:
```
jolyonb: Try removing the notification and webhook lines. We're using https://github.com/z4r/python-coveralls/tree/master to run coveralls, and it has parallel built in (see readme file at that repo).

jolyonb:
Ok, that didn't work. I need to head out, but perhaps this will be helpful? https://github.com/lemurheavy/coveralls-public/issues/1224

.travis.yml
ChristopherChudzicki: I've been looking at that and related threads for a while 🙍‍♂ 

This seems promising but more complicated: https://github.com/travis-ci/beta-features/issues/42

Trouble with just using the `combine` command from `coverage` is that it needs files written by `pytest`, which aren't shared between Travis stages, so they need to be uploaded somewhere (AWS S3 in the linked issue.)

ChristopherChudzicki:
I've been looking at that and related threads for a while 🙍‍♂

This seems promising but more complicated: travis-ci/beta-features#42

Trouble with just using the combine command from coverage is that it needs files written by pytest, which aren't shared between Travis stages, so they need to be uploaded somewhere (AWS S3 in the linked issue.)

PR #227, opened 2019-07-03
try codecov

ChristopherChudzicki:


ChristopherChudzicki:
So, I propose we use codecov. 

I tried this because 
 - in reading about merging reports on coveralls, a few people commented that they just switched to codecov
 - looking at https://github.com/codecov/codecov-python, it says "Reports are automatically combined with no extra setup. Each build is stored separately and combined." that made me happy 😄 

Definitely try clicking "[Continue to review full report at Codecov](https://codecov.io/gh/mitodl/mitx-grading-library/pull/227?src=pr&el=continue)"

jolyonb:
Good enough for me!

PR #228, opened 2019-07-03
Drop cov

ChristopherChudzicki:
Just testing that codecov shows a drop in coverage and it's easy to find where from.

(Actually, not a drop, since codecov has no data for master, but at least an uncovered line.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,11 @@ matrix:
       - "2.7"
     install:
       - pip install -r requirements.txt
-      - pip install python-coveralls
+      - pip install codecov
     script:
       - pytest --cov=mitxgraders --cov-report=term-missing
     after_success:
-      - coveralls
+      - codecov
 
   - language: python
     python:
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki:
I'm confused about why the bot didn't comment here, but the report did get generated:

https://codecov.io/gh/mitodl/mitx-grading-library/commit/9d6151f6fbf1d47d59dfe0562de28bab13aa5b64

ChristopherChudzicki:
Codecov seems not to recognize that the build passed...

Screen Shot 2019-07-03 at 5 04 12 PM

Well, I'll leave it for a while.

jolyonb:
Ok, not sure what happened here. Try again in a new PR, now that the base branch has coverage?

jolyonb:
Also, badges like in the following would be nice :-) https://github.com/TechnionYP5777/SmartCity-Market/wiki/Integrating-Codecov-with-a-GitHub-project-(Coverage-tool)

PR #233, opened 2019-07-05
Adding cosmetic replacement function to javascript

jolyonb:
This one is pretty simple. I found that some javascript replacements I was doing were interfering with wrapping, so I needed a separate function to run after wrapping occurred.

### DIFF
--- a/MJxPrep.js
+++ b/MJxPrep.js
@@ -98,6 +98,9 @@ if (window.MJxPrep) {
       return 'delta_{' + args[0] + ',' + args[1] + '}'
     })
 
+    // Do any custom cosmetic replacements (see end of script)
+    eqn = customCosmeticReplacements(eqn)
+
     // Return the preprocessed equation
     return eqn;
   }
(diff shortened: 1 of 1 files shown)


MJxPrep.js
```diff
@@ -627,8 +630,30 @@ if (window.MJxPrep) {
     funcToPostfix: funcToPostfix
   }
 
-  // A function to allow for custom replacements
   function customReplacements(eqn) {
+    /*
+     * A function to allow for custom replacements.
+     * This function is called *before* wrapping occurs, and should
+     * be used for implementing new functions, but not for doing
+     * cosmetic replacements. The output of this function must still
```
ChristopherChudzicki: I'm confused about what you mean by "cosmetic" change. Isn't everything AsciiMath/MathJax-related cosmetic? 

Is
```js
// Replace trans(...) -> {:(...)^T:}, with parentheses added as necessary
    eqn = replaceFunctionCalls(eqn, 'trans', funcToPostfix('^T') )
```
implementing a new function, or cosmetic?

jolyonb: The issue is that our wrapping function has regex that expects things to be syntactically equivalent to the grading library. So, when doing cosmetic changes, they need to be done after the wrapping is complete. I do agree that many of our function implementations are cosmetic... Do you think we can just do wrapping first?

ChristopherChudzicki: Well, it's better if `wrapFuncCalls` comes after anything that replaces traditional function calls `func(args)` with anything else, e.g., postfix notation:

```js
eqn = 'trans(A)'
eqn = wrapVariables(eqn)
eqn = wrapFuncCalls(eqn)
eqn = replaceFunctionCalls(eqn, 'trans', funcToPostfix('^T') )
console.log(eqn)
// {:{:({:A:})^T:}:} // has an extra set of parens and redundant outer `{:...:}`

eqn = 'trans(A)'
eqn = replaceFunctionCalls(eqn, 'trans', funcToPostfix('^T') )
eqn = wrapVariables(eqn)
eqn = wrapFuncCalls(eqn)
console.log(eqn)
// {:{:A:}^{:T:}:}
```
If I move the func/var wrapping after all `replaceFuncCalls` replacements, all of our tests still pass. 

Proposal: offer two custom replacement hooks,
 - `customPreReplacements(eqn)` called before ALL of our replacements except for whitespace cleaning
 - `customPostReplacements(eqn)` called after ALL of our default replacements

I think this is also easier to explain in documentation comments, and leads to more predictable values of the `eqn` argument.

jolyonb: Sounds good to me. Want me to implement it in this PR?

ChristopherChudzicki: Sure.

jolyonb:
Ok, fixed and ready for re-review.

PR #234, opened 2019-07-05
Introducting instructor variables

### LINKED ISSUE #160: Enable some way to turn off default constants
What it says on the tin.

jolyonb:
This introduces instructor variables, which can be used to remove a variable for use in student evaluations. This is useful when a variable is needed for DependentSampler, and can also be used to remove predefined constants.

This PR also prettifies some FormulaGrader error messages by adding quotation marks.

Resolves #160

### DIFF
--- a/mitxgraders/formulagrader/formulagrader.py
+++ b/mitxgraders/formulagrader/formulagrader.py
@@ -420,6 +420,11 @@ class FormulaGrader(ItemGrader):
                     should return True, False, 'partial', or a dictionary with required key
                     'grade_decimal' and optional key 'msg'. Comparer messages are ignored
                     when comparison succeeds (result['ok'] is True).
+
+        instructor_vars ([str]): A list of variable/constant names that cannot be used by
+            students. This can be useful in constructing DependentSampler expressions or
+            blacklisting constants. Note that this list is not validated against the list
+            of constants/variables.
     """
 
     default_functions = DEFAULT_FUNCTIONS.copy()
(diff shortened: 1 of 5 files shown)


mitxgraders/formulagrader/formulagrader.py
```diff
@@ -420,6 +420,11 @@ class FormulaGrader(ItemGrader):
                     should return True, False, 'partial', or a dictionary with required key
                     'grade_decimal' and optional key 'msg'. Comparer messages are ignored
                     when comparison succeeds (result['ok'] is True).
+
+        instructor_vars ([str]): A list of variable/constant names that cannot be used by
+            students. This can be useful in constructing DependentSampler expressions or
+            blacklisting constants. Note that this list is not validated against the list
```
ChristopherChudzicki: Idea: could we just make `blacklist` accept (func names, var names, constant names) instead of just function names?

(I haven't though about this too much from an implementation point of view.)

jolyonb: I don't want to do that, because it's possible to have functions and variables with the same name, and blacklist just takes in a list of strings. It would be impossible to distinguish between them, no?

jolyonb:
Do you think checking the `EvalMetaData` tuple would be a better way of implementing this?

mitxgraders/formulagrader/formulagrader.py
ChristopherChudzicki: > because it's possible to have functions and variables with the same name

Good point. I wish we had called it `blacklist_funcs` and `blacklist_vars`.

Anyway, `instructor_vars` is the same goal as blacklist, just for vars instead of funcs. What do you think of calling it `blacklist_vars`?

The only disadvantage is that its for vars and consts, but `instructor_vars` has that disadvantage, too.

ChristopherChudzicki:
> Do you think checking the EvalMetaData tuple would be a better way of implementing this?

I don't think it matters. It's slightly more consistent with how we hand blacklist, but that's no big deal and it would require changing a few intermediate funcs. I think fine to leave it.

mitxgraders/formulagrader/formulagrader.py
ChristopherChudzicki: Actually, insturctor_vars is good. I'm just gonna merge.

jolyonb: Hah, did you decide that it wasn't quite the same thing as a blacklist?

PR #235, opened 2019-07-05
Allowing DiscreteSet to work with MathArrays

### LINKED ISSUE #184: DiscreteSet only works for a set of numbers, not MathArrays
As noted in #182, we should expand the `DiscreteSet` sampler to allow for sampling a discrete set of `MathArray` objects.

jolyonb:
Pretty much what it says on the tin. Just needed to allow for a more general voluptuous schema.

Resolves #184


ChristopherChudzicki:
👍

PR #239, opened 2019-07-06
Removing pylint from requirements.txt

jolyonb:
I tried setting up a fresh install of this repo in python 2 and 3. The python 2 version is failing because pylint doesn't install properly. We're not actually using it, so I'm removing it from requirements.txt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,9 +4,8 @@ pyparsing==2.0.7
 scipy==0.14.0
 six==1.11.0
 
-# For testing/linting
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki:
OK, but `pylint` is great and you should set it up for your editor! But yeah, our CI doesn't use pylint, so fine to remove it.

jolyonb:
Yes, I've got pylint going in sublime, never fear ;-)

PR #248, opened 2019-07-06
Setting up attempt-based grading

### LINKED ISSUE #137: Implement options for grading based on attempt number
We finally have the capability! I've finally gotten edX to merge [this PR](https://github.com/edx/edx-platform/pull/18149), which allows an author to request the attempt number be passed to the check function. It's straightforward to see if the attempt number kwarg arrived at the grader, and I think we can apply a simple scheme to decrease credit based on attempt number:

* Start decreasing points awarded from attempt number X >= 2
* Minimum points awarded Y with 0 < Y <= 1
* Number of steps N until Y points are awarded

Then just perform a linear interpolation. A fancy feature would be to have a defaults file for this, so you can change the settings quickly over an entire course. (Maybe in the plugins folder?)

jolyonb:
This PR introduces the ability to have decreasing credit on increasing attempt number. See [this PR](https://github.com/edx/edx-platform/pull/18149) at edX which introduces the required feature and gives an example of how to use it.

I decided against including a feature to have a plugin set course-wide defaults for this, as that's a separate feature that could be used to set defaults for any option in any grader. I've made a new issue for such a feature.

Resolves #137

### DIFF
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -225,9 +225,10 @@ class AbstractGrader(ObjectWithSchema):
         Mutates result directly.
         """
         if attempt_number is None:
-            self.log("Attempt-based credit requested, but attempt number "
-                     "not passed through to grader")
-            return
+            msg = ("Attempt number not passed to grader as keyword argument 'attempt'. "
+                   'The attribute <code>cfn_extra_args="attempt"</code> may need to be '
+                   "set in the <code>customresponse</code> tag.")
+            raise ConfigError(msg)
         self.log("Attempt number {}".format(attempt_number))
 
         # How far past the point of decreasing credit are we?
@@ -254,13 +255,13 @@ class AbstractGrader(ObjectWithSchema):
                 if results_dict['grade_decimal'] > 0:
                     grade = results_dict['grade_decimal'] * credit
                     results_dict['grade_decimal'] = grade
-                    results_dict['ok'] = {0: False, 1: True}.get(grade, 'partial')
+                    results_dict['ok'] = self.grade_decimal_to_ok(grade)
                     changed_result = True
         else:
             if result['grade_decimal'] > 0:
                 grade = result['grade_decimal'] * credit
                 result['grade_decimal'] = grade
-                result['ok'] = {0: False, 1: True}.get(grade, 'partial')
+                result['ok'] = self.grade_decimal_to_ok(grade)
                 changed_result = True
 
         # Append the message if credit was reduced
@@ -273,6 +274,11 @@ class AbstractGrader(ObjectWithSchema):
             if result[key]:
                 result[key] += '\n\n'
             result[key] += msg.format(attempt_number, credit)
+
+    @staticmethod
+    def grade_decimal_to_ok(grade):
+        """Converts a grade decimal into an 'ok' value: True, False or 'partial'"""
+        return {0: False, 1: True}.get(grade, 'partial')
 
     @staticmethod
     def format_messages(result):
@@ -447,11 +453,6 @@ class ItemGrader(AbstractGrader):
         Usually this is a just a string.
         """
         return Schema(text_string)(expect)
-
-    @staticmethod
-    def grade_decimal_to_ok(grade):
-        """Converts a grade decimal into an 'ok' value: True, False or 'partial'"""
-        return {0: False, 1: True}.get(grade, 'partial')
 
     @staticmethod
     def standardize_cfn_return(value):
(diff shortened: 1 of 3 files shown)


jolyonb:
@ChristopherChudzicki Ready for review.

mitxgraders/baseclasses.py
```diff
@@ -169,6 +197,10 @@ def __call__(self, expect, student_input):
                     formatted = msg.format(student_input)
                 raise StudentFacingError(formatted)
 
+        # Handle partial credit based on attempt number
+        if self.config['attempt_based_credit']:
+            self.apply_attempt_based_credit(result, kwargs.get('attempt'))
```
jolyonb: This time I have a justification for my single-use well-named function: somebody may want to shadow it!

jolyonb:

ChristopherChudzicki:
This happened to me once... the report did generate: 

https://codecov.io/gh/mitodl/mitx-grading-library/pull/248

But codecov seems not to realize that the build passed, which dissuades the bot from commenting, I believe:

https://codecov.io/gh/mitodl/mitx-grading-library/commits

You can see the absence of a green check mark

mitxgraders/baseclasses.py
```diff
@@ -187,6 +219,61 @@ def __call__(self, expect, student_input):
...
+        self.log("Maximum credit is {}".format(credit))
+
+        # Multiply all grades by credit, updating from 'ok'=True to 'partial' as needed
+        changed_result = False
+        if "input_list" in result:
+            for results_dict in result['input_list']:
+                if results_dict['grade_decimal'] > 0:
+                    grade = results_dict['grade_decimal'] * credit
+                    results_dict['grade_decimal'] = grade
+                    results_dict['ok'] = {0: False, 1: True}.get(grade, 'partial')
```
ChristopherChudzicki: ItemGrader has 

```python
@staticmethod
def grade_decimal_to_ok(grade):
    """Converts a grade decimal into an 'ok' value: True, False or 'partial'"""
    return {0: False, 1: True}.get(grade, 'partial')
```
you could move that into Abstract grader and use it here, if you want.

tests/test_base.py
```diff
@@ -336,3 +336,211 @@ def test_ensure_text_inputs_errors():
...
+    assert grader(None, 'dog', attempt=2) == expected_result
+
+    expected_result = {'msg': 'Maximum credit for attempt #3 is 0.00.', 'grade_decimal': 0, 'ok': False}
+    assert grader(None, 'cat', attempt=3) == expected_result
+    expected_result = {'msg': '', 'grade_decimal': 0, 'ok': False}
+    assert grader(None, 'dog', attempt=3) == expected_result
+
+    # Ensure that messages are appended as appropriate
+    grader = StringGrader(
+        answers={'expect': 'cat', 'msg': 'Meow!'},
```
ChristopherChudzicki: Do you want to change this to `answers={'expect': 'cat', 'grade_decimal': 0.5, 'msg': 'Meow!'},` just to asssert somewhere that this does work as expected on partial-credit answers?

(It does work this way, just not asserted in a test unless I missed it.)

tests/test_base.py
```diff
@@ -336,3 +336,211 @@ def test_ensure_text_inputs_errors():
...
+    assert grader(None, 'cat', attempt=2) == expected_result
+    expected_result = {'msg': 'too bad', 'grade_decimal': 0, 'ok': False}
+    assert grader(None, 'dog', attempt=2) == expected_result
+
+    expected_result = {'msg': 'Meow!<br/>\n<br/>\nMaximum credit for attempt #3 is 0.00.', 'grade_decimal': 0, 'ok': False}
+    assert grader(None, 'cat', attempt=3) == expected_result
+    expected_result = {'msg': 'too bad', 'grade_decimal': 0, 'ok': False}
+    assert grader(None, 'dog', attempt=3) == expected_result
+
+    # Ensure that things behave when an attempt number is not passed in
```
ChristopherChudzicki: Is the behavior below really correct? Or should we raise a `ConfigError` to the effect of

"Attempt number not passed to grader as keyword argument `'attempt'`. The input tag attribute `cfn_extra_args="attempt"` may need to be set.

(This is a question, since I've never written an attempt-based problem.)

ChristopherChudzicki:
I wish that "Request changes" review button was "Request changes or answers to questions..." 😄

mitxgraders/baseclasses.py
jolyonb: Good point. Done.

tests/test_base.py
jolyonb: Done.

tests/test_base.py
jolyonb: Hmm. If the attempt number isn't passed in, I was using that as a mechanism for cheaply turning this feature off (useful when combined with the defaults). You're correct in that it's presently implemented as a "fail silently". Do you think its worth having a config error raised instead so that people know there's a second piece that they need to implement? Eh, seems reasonable to me. I'm going to do it.

jolyonb:
Updated and ready for final review.

PR #251, opened 2019-07-08
Adding instructor variables back in to debug output

jolyonb:
This fixes a small issue with the implementation of instructor variables: they need to be added back in to the debug output. I haven't made a test to fix this, because it's one of those functiongrader debug log things (I was going to co-opt the corresponding sibling variable test, but found that there wasn't one for that either!).


jolyonb:

PR #252, opened 2019-07-08
Infer answers from expect values for SingleListGrader

jolyonb:
This PR generalizes the implementation of reading the `expect` keyword to infer answers. In particular, it does the following:

* Makes it explicit that once an answer is inferred, that answer is stored in the configuration
* Ensures that when answers are being inferred, newly-provided `expect` parameters override the old ones
* Adds the inferred answers to the debug log
* Generalizes the debug log implementation slightly to provide access to it before the `AbstractGrader.__call__` function is called, while ensuring that the log is cleared between calls to a grader
* Implements answer inference as a shadowable function
* Implements answer inference for `SingleListGrader`
* Implements proper delimiter checking across nested `SingleListGrader`s

### DIFF
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -12,6 +12,7 @@ import numbers
 import abc
 import pprint
 import six
+import json
 from voluptuous import Schema, Required, All, Any, Range, MultipleInvalid
 from voluptuous.humanize import validate_with_humanized_errors as voluptuous_validate
 from mitxgraders.version import __version__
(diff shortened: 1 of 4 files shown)


mitxgraders/listgrader.py
```diff
@@ -644,6 +644,17 @@ def __init__(self, config=None, **kwargs):
         # Step 2: Validate the answers
         self.config['answers'] = self.schema_answers(self.config['answers'])
 
+        # Step 3: Ensure that nested SingleListGraders all use different delimiters
```
jolyonb: This validation wasn't being performed in the correct place previously.

jolyonb:
Also, I found a really neat way to avoid the unicode `u` issue by using json output!

jolyonb:

ChristopherChudzicki:
@jolyonb Can you resolve merge conflicts on this before review? Sorry for the review backlog =/

jolyonb:
Rebased and resolved merge conflicts.

jolyonb:
Ugh, I pushed a wrong branch here... Sorry for the extra messages. Should be all fixed now.

mitxgraders/baseclasses.py
```diff
@@ -629,16 +646,47 @@ def check_response(self, answer, student_input, **kwargs):
...
         The same as AbstractGrader.__call__, except that we try to infer
         answers from expect argument if answers are not specified in the
-        grader configuration.
+        grader configuration. The actual inference is done by infer_answers,
+        which can be shadowed.
+        """
+        # If expect is provided, infer an answer if we either don't have an answer or
+        # are always inferring answers
+        if expect is not None and (self.inferring_answers or not self.config['answers']):
+            self.config['answers'] = self.infer_answers(expect)
```
ChristopherChudzicki: Two minor comments:

First: There's no reason to store this in `self.config['answers']` yet, right? The reason I ask is that I believe in all other parts of the library, `self.config` only ever stores validated data and `self.infer_answers(expect)` has not yet been validated. (Admittedly, it will be validated a few lines from now).

Second: At least in my mind, there is an internal distinction between an "answer" and an "expect value". `infer_answers` is actually inferring the expect value, and `schema_answers` is inferring the answer value from the expect value. Do you think it's worth making a naming change?

jolyonb: Both reasonable. Addressed in commit.

PR #254, opened 2019-07-08
Ensure answers provided to SingleListGrader are nonempty

jolyonb:
This validates that the configuration for `SingleListGrader` does not have any empty answers, if such answers would require the student to enter an answer that would run afoul of the `missing_error` setting.

### DIFF
--- a/mitxgraders/listgrader.py
+++ b/mitxgraders/listgrader.py
@@ -9,6 +9,7 @@ Both work by farming out the individual objects to other graders.
 """
 from __future__ import print_function, division, absolute_import, unicode_literals
 
+import six
 import numpy as np
 from voluptuous import Required, Any
 from mitxgraders.helpers import munkres
@@ -704,6 +705,12 @@ class SingleListGrader(ItemGrader):
             if len(answer_list) != len(answers_tuple[0]):
                 raise ConfigError("All possible list answers must have the same length")
 
+        # Check for empty entries anywhere in answers_tuple (which can be a nested mess!)
+        # We do this before validating answer_list, as strings may be coerced into other
+        # objects by schema validation (e.g., FormulaGrader coerces expect into a dict)
+        if self.config['missing_error']:
+            demand_no_empty(answers_tuple)
+
         # Validate answer_list using the subgrader
         for answer_list in answers_tuple:
             for index, answer in enumerate(answer_list):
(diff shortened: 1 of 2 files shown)


mitxgraders/listgrader.py
```diff
@@ -682,6 +683,12 @@ def schema_answers(self, answer_tuple):
             if self.config['delimiter'] == subgrader.config['delimiter']:
                 raise ConfigError("Nested SingleListGraders must use different delimiters.")
 
+        # Check for empty entries anywhere in answers_tuple (which can be a nested mess!)
+        # We do this before validating answer_list, as strings may be coerced into other
+        # objects by schema validation (e.g., FormulaGrader coerces expect into a dict)
```
ChristopherChudzicki: 👍

mitxgraders/listgrader.py
```diff
@@ -734,3 +741,33 @@ def check_response(self, answer, student_input, **kwargs):
...
+            demand_no_nonempty(item)
+    else:
+        test = obj
+        if isinstance(test, dict):
+            if 'expect' in test:
+                test = test['expect']
+            else:
+                return
+        # We're doing this BEFORE validation, so could be given text or a string
+        if isinstance(test, six.text_type) or isinstance(test, six.string_types):
```
ChristopherChudzicki: You only need `isinstance(test, six.string_types)` here ( `(basestring,)` in 2, `(str,)` in 3)

mitxgraders/listgrader.py
```diff
@@ -734,3 +741,33 @@ def check_response(self, answer, student_input, **kwargs):
                                         partial_credit=self.config['partial_credit'])
 
         return result
+
+def demand_no_nonempty(obj):
```
ChristopherChudzicki: should this be `demand_all_nonempty`? or `demand_no_empty`?

mitxgraders/listgrader.py
jolyonb: Happy to make the change, but if someone imports unicode literals, then only doing `string_types` will fail though, right?

mitxgraders/listgrader.py
jolyonb: Oops! Fixed.

mitxgraders/listgrader.py
ChristopherChudzicki: Should work... `basestring` is the superclass of string and unicode.

```python
import six

# Python 2.7
isinstance(u'cat', six.string_types) # True
isinstance('cat', six.string_types) # True
```

jolyonb: Good point. Now fixed.

jolyonb:
Rebase coming up...

PR #255, opened 2019-07-08
Matrix partial credit

ChristopherChudzicki:
Adds MatrixEntryComparer, a configurable grader that implements partial credit for matrices based on entry.

### DIFF
--- a/mitxgraders/comparers/comparers.py
+++ b/mitxgraders/comparers/comparers.py
@@ -161,7 +161,6 @@ class MatrixEntryComparer(CorrelatedComparer):
 
     Configuration
     =============
-
         transform (None | function): same as EqualityComparer (default None)
         entry_partial_credit ('proportional' | number): Determines how partial credit
             is awarded. If set to 'proportional', then credit is proportional to
@@ -173,6 +172,7 @@ class MatrixEntryComparer(CorrelatedComparer):
         entry_partial_msg (str): A text string message shown when partial credit
             is awarded. The string may optionally contain the formatting key {error_indices},
             which will be replaced with the indices of the incorrect matrix entries.
+            To show no message, use the the empty string.
             Default value is:
             "Some matrix entries are incorrect, marked below:\n{error_locations}"
     """
@@ -191,15 +191,17 @@ class MatrixEntryComparer(CorrelatedComparer):
 
         Arguments:
             format_string: a string that may contain {error_locations} formatting key.
-            locs: a boolean error with False values indicating errors
+            locs: a boolean array with False values indicating incorrect entries
         """
         # Not the most elegant way to do these replacements, but this was what
-        # I came up with the minimize the amount of extra u prefixes in Python 2
-        bad_str = '\u2717'
-        good_str = '\u2713'
+        # I came up with to minimize the amount of extra u prefixes in Python 2
+        bad_str = '<span style="color:#b20610">\u2717</span>'
+        good_str = '<span style="color:#008100">\u2713</span>'
         matrix_as_text = six.text_type(locs).replace("  ", " ").replace("[ ", "[")
         matrix_as_text = matrix_as_text.replace("True", good_str).replace("False", bad_str)
-        return format_string.format(error_locations=matrix_as_text)
+        matrix_as_text = matrix_as_text.replace('\n', '<br/>')
+        formatted_locs = '<pre>{mat}</pre>'.format(mat=matrix_as_text)
+        return format_string.format(error_locations=formatted_locs)
 
     @staticmethod
     def validate(expected_evals, student_evals, utils):
--- a/mitxgraders/comparers/comparers.py
+++ b/mitxgraders/comparers/comparers.py
@@ -195,6 +195,8 @@ class MatrixEntryComparer(CorrelatedComparer):
         """
         # Not the most elegant way to do these replacements, but this was what
         # I came up with to minimize the amount of extra u prefixes in Python 2
+
+        # These are the edX colors, at least as of July 2019
         bad_str = '<span style="color:#b20610">\u2717</span>'
         good_str = '<span style="color:#008100">\u2713</span>'
         matrix_as_text = six.text_type(locs).replace("  ", " ").replace("[ ", "[")
--- a/mitxgraders/comparers/comparers.py
+++ b/mitxgraders/comparers/comparers.py
@@ -167,7 +167,7 @@ class MatrixEntryComparer(EqualityComparer):
             which will be replaced with the indices of the incorrect matrix entries.
     """
 
-    default_msg = "Matrix entries at [row column] locations {error_indices} are incorrect."
+    default_msg = "Matrix entries at indices {error_indices} are incorrect."
     schema_config = EqualityComparer.schema_config.extend({
         Required('entry_partial_credit', default=0): Any(All(Number, Range(0, 1)), 'proportional'),
         Required('entry_partial_msg', default=default_msg): text_string
--- a/mitxgraders/comparers/comparers.py
+++ b/mitxgraders/comparers/comparers.py
@@ -59,7 +59,15 @@ from mitxgraders.exceptions import InputTypeError, StudentFacingError
 from mitxgraders.helpers.validatorfuncs import is_callable, Nullable, text_string
 from mitxgraders.helpers.calc.mathfuncs import is_nearly_zero
 from mitxgraders.helpers.calc.math_array import are_same_length_vectors, is_vector
-from mitxgraders.comparers.baseclasses import Comparer
+from mitxgraders.comparers.baseclasses import Comparer, CorrelatedComparer
+
+def identity_transform(x):
+    """
+    Returns the input.
+
+    Note: used instead of lambdas because it prints a nice name.
+    """
+    return x
 
 class EqualityComparer(Comparer):
     """
@@ -120,7 +128,11 @@ class EqualityComparer(Comparer):
 
     """
     schema_config = Schema({
-        Required('transform', default=None): Nullable(is_callable)
+        Required('transform', default=None): All(
+            Nullable(is_callable),
+            # if f is None, coerce to identity function
+            lambda f: identity_transform if f is None else f
+        )
     })
 
     @staticmethod
@@ -130,23 +142,19 @@ class EqualityComparer(Comparer):
             shape = tuple() if isinstance(expected_eval, Number) else expected_eval.shape
             utils.validate_shape(student_eval, shape)
 
-    def transform(self, evaluation):
-        f = self.config['transform']
-        return evaluation if f is None else f(evaluation)
-
-
     def __call__(self, comparer_params_eval, student_eval, utils):
         expected_eval = comparer_params_eval[0]
         self.validate(expected_eval, student_eval, utils)
 
-        expected_eval = self.transform(expected_eval)
-        student_eval = self.transform(student_eval)
+        transform = self.config['transform']
+        expected_eval = transform(expected_eval)
+        student_eval = transform(student_eval)
 
         return utils.within_tolerance(expected_eval, student_eval)
 
 equality_comparer = EqualityComparer()
 
-class MatrixEntryComparer(EqualityComparer):
+class MatrixEntryComparer(CorrelatedComparer):
     """
     Default comparer for MatrixGrader. Compares student and instructor matrix
     evaluations entry-by-entry for equality.
@@ -186,20 +194,27 @@ class MatrixEntryComparer(EqualityComparer):
         locs_as_str = ", ".join(map(six.text_type, locs+1))
         return format_string.format(error_indices=locs_as_str)
 
-    def __call__(self, comparer_params_eval, student_eval, utils):
-        expected_eval = comparer_params_eval[0]
-        self.validate(expected_eval, student_eval, utils)
+    @staticmethod
+    def validate(expected_evals, student_evals, utils):
+        for x, y in zip(expected_evals, student_evals):
+            EqualityComparer.validate(x, y, utils)
 
-        num_entries = expected_eval.size
+    def __call__(self, comparer_params_evals, student_evals, utils):
+        expected_evals = [params[0] for params in comparer_params_evals]
+        self.validate(expected_evals, student_evals, utils)
 
-        expected_eval = self.transform(expected_eval)
-        student_eval = self.transform(student_eval)
-
+        transform = self.config['transform']
+        expected_evals = [transform(x) for x in expected_evals]
+        student_evals = [transform(x) for x in student_evals]
         vec_within_tol = np.vectorize(utils.within_tolerance)
-        # compared_items is a boolean array of entry-by-entry comparisons
-        compared_items = vec_within_tol(expected_eval, student_eval)
-        percent_correct = np.sum(compared_items).item()/num_entries
-        wrong_locs = np.argwhere(np.logical_not(compared_items))
+        # comparisons_by_eval is a boolean array of entry-by-entry comparisons,
+        # one for each comparison. Its numpy shape is (n_evals, *eval_shape)
+        comparisons_by_eval = vec_within_tol(expected_evals, student_evals)
+        comparisons_summary = np.all(comparisons_by_eval, axis=0)
+
+        num_entries = comparisons_summary.size
+        percent_correct = np.sum(comparisons_summary).item()/num_entries
+        wrong_locs = np.argwhere(np.logical_not(comparisons_summary))
         msg = self.format_message_with_locations(self.config['entry_partial_msg'], wrong_locs)
         partial_credit = self.config['entry_partial_credit']
 
--- a/mitxgraders/comparers/comparers.py
+++ b/mitxgraders/comparers/comparers.py
@@ -148,20 +148,20 @@ equality_comparer = EqualityComparer()
 
 class MatrixEntryComparer(EqualityComparer):
     """
-    Default comparer for MatrixGrader. Compares student and instructor matirx
+    Default comparer for MatrixGrader. Compares student and instructor matrix
     evaluations entry-by-entry for equality.
 
     Configuration
     =============
 
-        transform (None | function): same as EqualityComparer, defaults to None
-        entry_partial_credit ('proportional'|number): Determines how partial credit
+        transform (None | function): same as EqualityComparer (default None)
+        entry_partial_credit ('proportional' | number): Determines how partial credit
             is awarded. If set to 'proportional', then credit is proportional to
-            the number of correct matrix entries. If a numeric value is provided,
-            this flat rate of partial credit is provided as long as some but
-            not all entries are correct.
+            the number of correct matrix entries. If a numeric value betweem 0 and 1
+            is provided, this flat rate of partial credit is provided as long as
+            some but not all entries are correct. Default is the numeric value 0
+            (no partial credit).
 
-            Default is the numeric value 0 (no partial credit).
         entry_partial_msg (str): A text string message shown when partial credit
             is awarded. The string may optionally contain the formatting key {error_indices},
             which will be replaced with the indices of the incorrect matrix entries.
@@ -202,9 +202,6 @@ class MatrixEntryComparer(EqualityComparer):
         wrong_locs = np.argwhere(np.logical_not(compared_items))
         msg = self.format_message_with_locations(self.config['entry_partial_msg'], wrong_locs)
         partial_credit = self.config['entry_partial_credit']
-
-        if partial_credit == 0:
-            return percent_correct == 1
 
         if percent_correct == 0:
             return False
--- a/mitxgraders/comparers/comparers.py
+++ b/mitxgraders/comparers/comparers.py
@@ -52,10 +52,11 @@ from __future__ import print_function, division, absolute_import, unicode_litera
 from numbers import Number
 import numpy as np
 
-from voluptuous import Schema, Required
+import six
+from voluptuous import Schema, Required, Any, Range, All
 
 from mitxgraders.exceptions import InputTypeError, StudentFacingError
-from mitxgraders.helpers.validatorfuncs import is_callable, Nullable
+from mitxgraders.helpers.validatorfuncs import is_callable, Nullable, text_string
 from mitxgraders.helpers.calc.mathfuncs import is_nearly_zero
 from mitxgraders.helpers.calc.math_array import are_same_length_vectors, is_vector
 from mitxgraders.comparers.baseclasses import Comparer
@@ -102,7 +103,7 @@ class EqualityComparer(Comparer):
 
     The following example takes the norm of the expected answer and student input
     before comparison. Note the different method of changing the comparer.
-    >>> FormulaGrader.set_default_comparer(EqualityComparer(transform=np.linalg.norm))
+    >>> MatrixGrader.set_default_comparer(EqualityComparer(transform=np.linalg.norm))
     >>> grader = MatrixGrader(
     ...     answers='[1, 0, 0]'
     ... )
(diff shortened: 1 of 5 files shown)


mitxgraders/comparers/comparers.py
```diff
@@ -102,7 +103,7 @@ class EqualityComparer(Comparer):
 
     The following example takes the norm of the expected answer and student input
     before comparison. Note the different method of changing the comparer.
-    >>> FormulaGrader.set_default_comparer(EqualityComparer(transform=np.linalg.norm))
+    >>> MatrixGrader.set_default_comparer(EqualityComparer(transform=np.linalg.norm))
```
jolyonb: Oops, my bad!

ChristopherChudzicki: This no longer matters since I just reverted MatrixGrader.default_comparer to `equality_comparer`, but I still think it's nicer to use class override :)

ChristopherChudzicki:
Oops, I need one more test now that `MatrixGrader.default_comparer` is back to its original.

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
 
-        return utils.within_tolerance(expected_input, student_eval)
+    def __call__(self, comparer_params_eval, student_eval, utils):
+        expected_eval = comparer_params_eval[0]
+        self.validate(expected_eval, student_eval, utils)
+
+        expected_eval = self.transform(expected_eval)
+        student_eval = self.transform(student_eval)
+
+        return utils.within_tolerance(expected_eval, student_eval)
```
jolyonb: This is definitely cleaner :+1:

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
 
 equality_comparer = EqualityComparer()
 
+class MatrixEntryComparer(EqualityComparer):
+    """
+    Default comparer for MatrixGrader. Compares student and instructor matirx
```
jolyonb: Typo: matrix

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+    evaluations entry-by-entry for equality.
+
+    Configuration
+    =============
+
+        transform (None | function): same as EqualityComparer, defaults to None
```
jolyonb: Just bringing style in line with other such entries:
`Same as EqualityComparer (default None)`
Blank line afterwards.

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+        entry_partial_credit ('proportional'|number): Determines how partial credit
```
jolyonb: `('proportional' | number)`

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+            is awarded. If set to 'proportional', then credit is proportional to
+            the number of correct matrix entries. If a numeric value is provided,
```
jolyonb: numeric value between 0 and 1 (inclusive)

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+            this flat rate of partial credit is provided as long as some but
+            not all entries are correct.
```
jolyonb: Continue on this line:
`(default 0, for no partial credit)`
Blank line afterwards.

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+
+            Default is the numeric value 0 (no partial credit).
+        entry_partial_msg (str): A text string message shown when partial credit
+            is awarded. The string may optionally contain the formatting key {error_indices},
```
jolyonb: is awarded (even if that credit is 0).

Can we have a default on this? Something like `The following entries are correct: {error_indices}`

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+            which will be replaced with the indices of the incorrect matrix entries.
+    """
+
+    default_msg = "Matrix entries at indices {error_indices} are incorrect."
+    schema_config = EqualityComparer.schema_config.extend({
+        Required('entry_partial_credit', default=0): Any(All(Number, Range(0, 1)), 'proportional'),
```
jolyonb: Huh, I tried doing this somewhere else, but I used `All(float, Range(0, 1))`. It wouldn't accept 0 or 1, so I had to add them in separately in the `Any`. Guess I should've used `Number`?

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+        """
+        Returns format_string with {error_indices} replaced by 1-indexed error
+        locations.
+
+        Arguments:
+            format_string: a string that may contain {error_indices} formatting key.
+            locs: a 0-indexed numpy array of error locations
+        """
+        locs_as_str = ", ".join(map(six.text_type, locs+1))
+        return format_string.format(error_indices=locs_as_str)
```
jolyonb: What I'd really like to do here is to give a little map, rather than just indices. Something that outputs like the following:
```
[[x, x, ., x],
 [., ., x, .]]
```

Bonus points for making the x's red.

I think this shouldn't be too hard; construct a numpy array of booleans and get the whole thing formatted, then convert True and False appropriately?

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+
+        vec_within_tol = np.vectorize(utils.within_tolerance)
+        # compared_items is a boolean array of entry-by-entry comparisons
+        compared_items = vec_within_tol(expected_eval, student_eval)
+        percent_correct = np.sum(compared_items).item()/num_entries
+        wrong_locs = np.argwhere(np.logical_not(compared_items))
+        msg = self.format_message_with_locations(self.config['entry_partial_msg'], wrong_locs)
+        partial_credit = self.config['entry_partial_credit']
+
+        if partial_credit == 0:
```
jolyonb: This case is already handled by `EqualityComparer`. When `partial_credit == 0`, we should still be able to report the message.

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+            Default is the numeric value 0 (no partial credit).
+        entry_partial_msg (str): A text string message shown when partial credit
+            is awarded. The string may optionally contain the formatting key {error_indices},
+            which will be replaced with the indices of the incorrect matrix entries.
+    """
+
+    default_msg = "Matrix entries at indices {error_indices} are incorrect."
+    schema_config = EqualityComparer.schema_config.extend({
+        Required('entry_partial_credit', default=0): Any(All(Number, Range(0, 1)), 'proportional'),
+        Required('entry_partial_msg', default=default_msg): text_string
```
jolyonb: This should be `Nullable` so that the message can be turned off.

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+
+        transform (None | function): same as EqualityComparer, defaults to None
+        entry_partial_credit ('proportional'|number): Determines how partial credit
+            is awarded. If set to 'proportional', then credit is proportional to
+            the number of correct matrix entries. If a numeric value is provided,
+            this flat rate of partial credit is provided as long as some but
+            not all entries are correct.
+
+            Default is the numeric value 0 (no partial credit).
+        entry_partial_msg (str): A text string message shown when partial credit
```
jolyonb: Should be `(None | str)`.

mitxgraders/comparers/comparers.py
jolyonb: I think you can just remove these two lines.

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+
+        expected_eval = self.transform(expected_eval)
+        student_eval = self.transform(student_eval)
+
+        vec_within_tol = np.vectorize(utils.within_tolerance)
+        # compared_items is a boolean array of entry-by-entry comparisons
+        compared_items = vec_within_tol(expected_eval, student_eval)
+        percent_correct = np.sum(compared_items).item()/num_entries
+        wrong_locs = np.argwhere(np.logical_not(compared_items))
+        msg = self.format_message_with_locations(self.config['entry_partial_msg'], wrong_locs)
```
jolyonb: To account for `entry_partial_msg is None`, just put in an if statement here that otherwise sets `msg = ''`.

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -59,8 +60,23 @@ class MatrixGrader(FormulaGrader):
                         received objects is revealed.
 
         suppress_matrix_messages (bool): If True, suppresses all matrix-related
-            error messages from being displayed. Overrides shape_errors=True and
+            erro2r messages from being displayed. Overrides shape_errors=True and
```
jolyonb: Uh, I think that `2` was a keybump...

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +113,108 @@ class EqualityComparer(Comparer):
...
+        locations.
+
+        Arguments:
+            format_string: a string that may contain {error_indices} formatting key.
+            locs: a 0-indexed numpy array of error locations
+        """
+        locs_as_str = ", ".join(map(six.text_type, locs+1))
+        return format_string.format(error_indices=locs_as_str)
+
+    def __call__(self, comparer_params_eval, student_eval, utils):
```
jolyonb: This needs to be done with a correlated comparer rather than a simple comparer that computes the truth table across all samples. Otherwise, it's quite possible that the percentage and message are incorrect (especially when using integer sampling and Kronecker delta functions).

mitxgraders/comparers/comparers.py
jolyonb: I'm thinking that this should be `None` if no partial credit is to be awarded (and no message displayed), so that `0` can have no partial credit awarded, but the message displayed.

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -59,8 +60,23 @@ class MatrixGrader(FormulaGrader):
...
             is_raised=True. Defaults to False.
+
+        Additionally, the following three configuration keys will be passed to
+        the default comparer and be used to specify partial credit options.
+
+        entry_partial_credit ('proportional'|number): Determines how partial credit
```
jolyonb: Rather than repeat all this text, I would just include the entry names and say "See MatrixEqualityComparer". The docstrings aren't for the end users of this library, so anybody who's diving into here can follow that chain.

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -59,8 +60,23 @@ class MatrixGrader(FormulaGrader):
...
+            is awarded. If set to 'proportional', then credit is proportional to
+            the number of correct matrix entries. If a numeric value is provided,
+            this flat rate of partial credit is provided as long as some but
+            not all entries are correct.
+
+            Default is the numeric value 0 (no partial credit).
+        entry_partial_msg (str): A text string message shown when partial credit
+            is awarded. The string may optionally contain the formatting key {error_indices},
+            which will be replaced with the indices of the incorrect matrix entries.
+        transform (None | function): same as EqualityComparer, defaults to None
```
jolyonb: I think you meant to leave off transform? At least, you did in the code below. I think it makes sense to leave off `transform`; anybody wanting to use that key should set up the comparer properly. This means that there's only two configuration keys (see above - `Additionally, the following three...`).

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -82,11 +98,29 @@ def schema_config(self):
...
+        comparer_config = {key: unvalidated_config[key]
+                           for key in keys if key in unvalidated_config}
+        if comparer_config:
+            return MatrixEntryComparer(comparer_config)
+        return cls.default_comparer
+
     def __init__(self, config=None, **kwargs):
+        # Set default_comparer on as an instance property
+        # This will be used by schema_config during validation to set defaults
+        unvalidated_config = config if config is not None else kwargs
```
jolyonb: Is there a reason not to do this after the `super` call? That way, the config is validated...

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -82,11 +98,29 @@ def schema_config(self):
             }): {
                 Required('is_raised', default=True): bool,
                 Required('msg_detail', default='type'): Any(None, 'type', 'shape')
-            }
+            },
+            Required('entry_partial_credit', default=0): Any(All(Number, Range(0, 1)), 'proportional'),
+            Required('entry_partial_msg', default=MatrixEntryComparer.default_msg): text_string
         })
 
+    @classmethod
```
jolyonb: I don't see why this needs to be an independent method... I'm also a little wary that this is essentially saying that `MatrixGrader` doesn't benefit from the use of `set_default_comparer`, because its default comparer is always set through the config... Perhaps that's why you went with unvalidated here? If so, then it's weird for those config values to have defaults... I think we should probably talk through this one.

tests/formulagrader/test_matrixgrader.py
```diff
@@ -264,3 +264,90 @@ def test_suppress_matrix_messages():
     # Note that we haven't suppressed all errors:
     with raises(ArgumentError):
         grader(None, 'sin(1, 2)')
+
+def test_default_grader_instance_override():
```
jolyonb: I'm not going to review the tests at this stage.

mitxgraders/comparers/comparers.py
ChristopherChudzicki: If you want no message, just supply `entry_partial_msg=''`. Or am I missing something?

Maybe `None` is slightly more readable? But we use `''` for this purpose elsewhere that's what we do for most other error messages, right?

mitxgraders/comparers/comparers.py
ChristopherChudzicki: Sure

mitxgraders/comparers/comparers.py
ChristopherChudzicki: See above.

mitxgraders/comparers/comparers.py
ChristopherChudzicki: Great idea! I may not get the bonus points, we'll see.

mitxgraders/comparers/comparers.py
jolyonb: Oh, obviously ;-)

mitxgraders/comparers/comparers.py
ChristopherChudzicki: sure

mitxgraders/formulagrader/matrixgrader.py
ChristopherChudzicki: `MatrixGrader` does benefit from `set_default_comparer` in the same way that `FormulaGrader` does. Note that `get_default_comparer` returns `cls.default_comparer` unless `entry_partial...` keys are used. 

The `entry_partial...` keys cause `self.default_grader` to be overridden, but only on a per-instance basis.

That said, I can see how the code looks misleading. I'll rename some things.

mitxgraders/formulagrader/matrixgrader.py
ChristopherChudzicki: The reason to do this before `super` is that the `super` call will use `self.default_comparer` during validation to provide a comparer function as part of `answers` (if a comparer function was not explicitly provided). 

This portion of the config is still validated—twice, actually. It's validated when `super(MatrixGrader, self).__init__` runs, and it is also validated when the comparer function instantiates .

mitxgraders/formulagrader/matrixgrader.py
ChristopherChudzicki: Actually, I'm going to think a bit more about other ways to do this. This works, but it does require messing with the config about before validation occurs, which I don't love.

mitxgraders/comparers/comparers.py
ChristopherChudzicki: Yeah, I saw where you did that. Both seem fine. Maybe it's worth making a validator `numeric_range(x, y)` just for this case.

I was a little suprised that just `Range(0, 1)` doesn't work, but it throws `TypeError`s instead of `ValueError`s. I guess it's still good to declare Number or float or int, though, since things besides numbers might be comparable to 0 and 1 in Python.

jolyonb:
I should add that I do like the idea of having partial credit for a MatrixGrader configurable through that MatrixGrader. It makes setting a default really easy!

Perhaps the default grader for MatrixGrader should just be MatrixEntryComparer, using settings that would make it essentially equivalent to EqualityComparer. That way, validation can go through as per normal, as the partial credit options don't affect validation at all.

ChristopherChudzicki:
# Thoughts re modifying the default comparer on an instance level

The issue is really that during when `voluptuous` runs its validator functions on `config`, the keys can't communicate. So it's hard for `config` keys `entry_partial_<whatever>` to influence `config['answers']` validation. 

In other instances where config key-values need "correlated" validation, we've performed additional validation steps after `super().__init__` runs. For example, `validate_no_collisions(self.config, keys=['variables', 'user_constants'])`. 

I don't think that's suitable in this case for two reasons: 

1. `schema_answers` will have already set the comparer functions. It bugs me to set the comparer to one thing, then to reset it again later to another thing.
2. More seriously, `schema_answers` is used by inherited `__call__` to coerce except values into answers when provided by the expect XML attribute. So `schema_answers` needs to be able to construct the correct comparer on its own.
    - (I guess FormulaGrader could shadow __call__ to make its own modifications to the expect value read from XML, but I'd rather not).

Modifying `default_comparer` on the instance level before `super().__init__` runs its voluptuous validation solves both of these problems. 

Moreover, I think it conceptually makes sense to modify the default comparer. Consider that in the situation below
```python
grader = MatrixGrader(
    answers = (
        'expect_value_a',
        {'comparer': comparer1, 'comparer_params': ["whatever params"]}
        'expect_value_b',
        {'comparer': comparer2, 'comparer_params': ["whatever params"]}
    ),
    entry_partial_credit='proportional'
)
```
the comparer `MatrixEntryComparer(entry_partial_credit=0.5)` should be used for `expect_value_a` and expect_value_b`, but not for the other two answers, and in this sense it is the default comparer for this instance of MatrixGrader.

jolyonb:
This sounds reasonable to me. I am wary however of the interaction of your proposal with `set_default_grader`. Do you have an idea of how you want to reconcile the two?

ChristopherChudzicki:
Well, `set_default_comparer` is a classmethod that sets the default for the class, not for the instance. So the only conflict I see is that a user might set a class-level default comparer of NEW_CLASS_DEFAULT **and** use the `entry_partial` keys, then be surprised when `MatrixEntryComparer` was used instead of NEW_CLASS_DEFAULT.

An alternative syntax I considered was to allow FormulaGrader and its subclasses to specify a top-level comparer in the configuration, that would be the instance-level default. So something like:

```python
grader = MatrixGrader(
    answer='A*B',
    variables=['A', 'B']
    comparer=MatrixEntryComparer(partial_credit=0.5)
)
```
where `config['comparer']` become the instance-level default.

The advantage to this approach would be (1) more general: good for FormulaGrader, too. (2) makes more clear that you're overriding the default comparer. But it also requires more direct interaction with comparers, which is a bit advanced, I suppose. I personally like this better. 

(To be clear, I'd still be setting `default_comparer` on the instance level.)

jolyonb:
Hmm. If anybody is wanting to play with comparers at any stage, they need to get their hands dirty. I think the `comparer` flag is cleaner than the `set_default_comparer` method we're currently using (and bonus, can be easily set as a course-wide default!). Up to you if you want to do this.

At the same time, being able to set the MatrixGrader comparer through simple options is also a really nice touch, because it makes it much more accessible. We can detect if those options were provided before overriding a comparer before validation (what you were proposing, I gather).

Ok, I consider this hashed out. I trust you :-)

mitxgraders/comparers/comparers.py
ChristopherChudzicki: > I'm thinking that this should be None if no partial credit is to be awarded (and no message displayed), so that 0 can have no partial credit awarded, but the message displayed.

I've crossed out my "Sure" above. 

If you want no partial credit and no message, just use `EqualityComparer`, which is the default.

jolyonb: Fair enough.

ChristopherChudzicki:
@jolyonb This is ready for re-review. Sorry for force push, I rebased but probably didn't need to.

I'm about to get on a flight to Seattle. If you can review in the next 5 or 6 hours, I'll try to address any comments on my next flight.

jolyonb:
Hey, no problem. I force push all the time, and enough stuff has been modified the past few days!

Have a good flight! There'll be a thorough review ready for you when you land.

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +121,118 @@ class EqualityComparer(Comparer):
...
+    default_msg = "Some matrix entries are incorrect, marked below:\n{error_locations}"
+    schema_config = EqualityComparer.schema_config.extend({
+        Required('entry_partial_credit', default=0): Any(All(Number, Range(0, 1)), 'proportional'),
+        Required('entry_partial_msg', default=default_msg): text_string
+    })
+
+    @staticmethod
+    def format_message_with_locations(format_string, locs):
+        """
+        Returns format_string with {error_locations} replaced by 1-indexed error
```
jolyonb: I think this is out-dated too.

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +121,118 @@ class EqualityComparer(Comparer):
...
+        locations.
+
+        Arguments:
+            format_string: a string that may contain {error_locations} formatting key.
+            locs: a boolean error with False values indicating errors
```
jolyonb: A boolean error or a boolean array?

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +121,118 @@ class EqualityComparer(Comparer):
...
+        """
+        # Not the most elegant way to do these replacements, but this was what
+        # I came up with the minimize the amount of extra u prefixes in Python 2
```
jolyonb: `the minimize` or `to minimize`?

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +121,118 @@ class EqualityComparer(Comparer):
...
+        bad_str = '\u2717'
+        good_str = '\u2713'
+        matrix_as_text = six.text_type(locs).replace("  ", " ").replace("[ ", "[")
+        matrix_as_text = matrix_as_text.replace("True", good_str).replace("False", bad_str)
```
jolyonb: I might wrap this in a `<code></code>` block to make sure that spacing looks ok. I'll test and get back to you.

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +121,118 @@ class EqualityComparer(Comparer):
...
+        return format_string.format(error_locations=matrix_as_text)
+
+    @staticmethod
+    def validate(expected_evals, student_evals, utils):
```
jolyonb: This is really cute!

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +121,118 @@ class EqualityComparer(Comparer):
...
+
+        entry_partial_msg (str): A text string message shown when partial credit
+            is awarded. The string may optionally contain the formatting key {error_indices},
+            which will be replaced with the indices of the incorrect matrix entries.
+            Default value is:
+            "Some matrix entries are incorrect, marked below:\n{error_locations}"
+    """
+
+    default_msg = "Some matrix entries are incorrect, marked below:\n{error_locations}"
+    schema_config = EqualityComparer.schema_config.extend({
```
jolyonb: That's a really nice way of building on the previous config without needing to subclass. Good job!

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +121,118 @@ class EqualityComparer(Comparer):
...
+
+        if percent_correct == 1:
+            return True
+        elif percent_correct == 0:
+            return {'ok': False, 'grade_decimal': 0, 'msg': msg}
+        elif partial_credit == 'proportional':
+            return {'ok': 'partial', 'grade_decimal': percent_correct, 'msg': msg}
+        else:
+            return {'ok': 'partial', 'grade_decimal': partial_credit, 'msg': msg}
+
```
jolyonb: This routine is beautiful!!! You should be proud of it!

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -82,11 +91,28 @@ def schema_config(self):
...
 
+    @staticmethod
+    def get_entry_comparer_config(unvalidated_config):
+        keys = ('entry_partial_credit', 'entry_partial_msg')
+        comparer_config = {key: unvalidated_config[key]
+                           for key in keys if key in unvalidated_config}
+        return comparer_config
+
     def __init__(self, config=None, **kwargs):
+        # Set default_comparer on as an instance property if entry_parial keys
```
jolyonb: Typo: `entry_partial`

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -82,11 +91,28 @@ def schema_config(self):
...
+        # are provided
+        unvalidated_config = config if config is not None else kwargs
+        entry_comparer_config = self.get_entry_comparer_config(unvalidated_config)
```
jolyonb: I propose removing the function call completely, and replacing with this:
```python
entry_comparer_config = {key: unvalidated_config[key]
                         for key in ('entry_partial_credit', 'entry_partial_msg')
                         if key in unvalidated_config}
```

mitxgraders/formulagrader/matrixgrader.py
jolyonb: Also, remove `on`

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -82,11 +91,28 @@ def schema_config(self):
...
+        if entry_comparer_config:
+            self.default_comparer = MatrixEntryComparer(entry_comparer_config)
+
         super(MatrixGrader, self).__init__(config, **kwargs)
+
```
jolyonb: I'm satisfied with this API and implementation.

tests/formulagrader/test_matrixgrader.py
```diff
@@ -264,3 +265,105 @@ def test_suppress_matrix_messages():
...
+        'ok': False,
+        'msg': six.u(
+            'Some matrix entries are incorrect, marked below:<br/>\n'
+            '[[{b} {b} {b}]<br/>\n'
+            ' [{b} {b} {b}]]'
+        ).format(g=GOOD, b=BAD),
+        'grade_decimal': 0
+    }
+
+GOOD = six.u('\u2713')
```
jolyonb: Can you move these above the previous function, which utilizes them?

tests/formulagrader/test_matrixgrader.py
```diff
@@ -264,3 +265,105 @@ def test_suppress_matrix_messages():
...
+    }
+    assert grader(None, '[[10, 20, 30], [40, 50, 60]]') == {
+        'ok': False,
+        'msg': six.u(
+            'Some matrix entries are incorrect, marked below:<br/>\n'
+            '[[{b} {b} {b}]<br/>\n'
+            ' [{b} {b} {b}]]'
+        ).format(g=GOOD, b=BAD),
+        'grade_decimal': 0
+    }
```
jolyonb: Can you add in a test for vectors (both "array" vectors and normal vectors) to ensure that they print properly? Can you also do a 3-tensor to ensure that prints properly too?

mitxgraders/comparers/comparers.py
```diff
@@ -112,30 +121,118 @@ class EqualityComparer(Comparer):
...
-        return utils.within_tolerance(expected_input, student_eval)
+        transform = self.config['transform']
+        expected_eval = transform(expected_eval)
+        student_eval = transform(student_eval)
+
+        return utils.within_tolerance(expected_eval, student_eval)
 
 equality_comparer = EqualityComparer()
 
+class MatrixEntryComparer(CorrelatedComparer):
```
jolyonb: I'm wondering if this should be `ArrayEntryComparer`, as it applies to vectors/matrices/tensors...

tests/formulagrader/test_matrixgrader.py
```diff
@@ -264,3 +265,105 @@ def test_suppress_matrix_messages():
...
+        entry_partial_msg=six.u('Partly correct, errors at\n{error_locations}')
+    )
+
+    assert grader(None, '[[1, 20], [30, 4]]')['msg'] == 'Partly correct'
+
+    formatted_msg = six.u(
+        'Partly correct, errors at<br/>\n'
+        '[[{g} {b}]<br/>\n'
+        ' [{b} {g}]]'
+    ).format(g=six.u('\u2713'), b=six.u('\u2717'))
```
jolyonb: Use `GOOD` and `BAD` here?

mitxgraders/comparers/comparers.py
jolyonb: So, I tested this out on edX. Code blocks don't work so well, but pre blocks do the job beautifully:
```html
<pre>[[✓ <span style='color:red;'>✗</span>]<br/> [<span style='color:red;'>✗</span> ✓]]</pre>
```
I added in the colors (just update your bad_str appropriately.

One thing to note is that the `<pre>` blocks need to have `\n` replaced with `<br/>` but not `<br/>\n` as is done by default, so you'd need to do the replacement here rather than rely on the default replacements we perform.

Will also require updating tests minimally.

mitxgraders/comparers/comparers.py
jolyonb: Of all my comments, this one is probably the most work lol ;-)

mitxgraders/formulagrader/matrixgrader.py
```diff
@@ -61,6 +62,14 @@ class MatrixGrader(FormulaGrader):
...
             error messages from being displayed. Overrides shape_errors=True and
             is_raised=True. Defaults to False.
+
+        Additionally, the configuration options
+            entry_partial_credit
+            entry_partial_msg
+        of MatrixEntryComparer can be passed directly to MatrixGrader to facilitate
+        partial credit without the explicit use of comparers. If either key
+        is included, MatrixEntryComparer is used as the default comparer for
+        that MatrixGrader instance with the given key values.
```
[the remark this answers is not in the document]
jolyonb: Or perhaps `equality_comparer`.

mitxgraders/comparers/comparers.py
jolyonb: Actually, I've changed my mind on this one. We have `MatrixGrader`, so `MatrixEntryComparer` is fine. Maybe just add to the docstring that this does any array, not just matrices ;-)

ChristopherChudzicki:
Thanks for the comments and feedback! I believe I've implemented everything as you suggested.

The vector/matrix feedback messages look good. Actually, tensor looks OK, too, but Asciimath doesn't render it well.

Screen Shot 2019-07-10 at 5 56 37 PM Screen Shot 2019-07-10 at 5 56 44 PM Screen Shot 2019-07-10 at 6 24 08 PM

About to take off, so won't be online for another few hours. (Our flight got delayed, so I did this in the layover...)

jolyonb:
Looks pretty good to me :) Just going through my comments step by step, but will likely merge soon.

mitxgraders/comparers/comparers.py
```diff
@@ -195,6 +195,8 @@ def format_message_with_locations(format_string, locs):
         """
         # Not the most elegant way to do these replacements, but this was what
         # I came up with to minimize the amount of extra u prefixes in Python 2
+
+        # These are the edX colors, at least as of July 2019
```
jolyonb: Nice touch!

PR #256, opened 2019-07-08
Option to turn off partial credit in ListGrader

### LINKED ISSUE #245: All or nothing in a ListGrader
Currently, ListGrader gives partial credit if you get some of the answers correct. It should be straightforward to turn off partial credit unless everything has a nonzero score (should probably add a warning that no partial credit is being awarded in this case).

jolyonb:
Implements a partial credit switch for `ListGrader` (which previously had partial credit always on). Works correctly in nested lists, so that partial credit can be turned off for a sublist, but left on for the outermost list.

Resolves #245


jolyonb:
Did you mean to merge this?

Issue #247, opened 2019-07-06
Expanded answer structures for SingleListGrader


jolyonb:
This follows from #3.

In order to give feedback on lists, I'm going to propose allowing for a dictionary to describe the answer for a list in the following way:
```python
answers = {'expect': [list of answers here], 'grade_decimal': 0.5, 'msg': 'Overall message here'}
```
The message would display only if all answers are correct; the `grade_decimal` value would multiply every box. One could also have tuples of such dictionaries. Just giving a list would be equivalent to having the expect keyword.

Note that this doesn't work nearly so nicely for ListGrader due to nested lists not having a nice place to put overall messages for sublists.

jolyonb:
Related to this is for a singlelistgrader to read its answer from the expect keyword.

ChristopherChudzicki:
Resolved, right?

jolyonb:
Should be all resolved :-)

ChristopherChudzicki:
Resolved by #261

PR #261, opened 2019-07-09
Overhauling SingleListGrader answer structure, validation and grading

jolyonb:
This PR updates the answer validation for SingleListGrader. It actually leverages `ItemGrader` in a neat way that makes everything nicer, in my opinion. Addresses #247 for SingleListGrader (but not ListGrader, which is a whole 'nother beast!).

The code:
```python
grader = SingleListGrader(
    answers=['a', 'b'],
    subgrader=StringGrader()
)
pprint.pprint(grader.config['answers'])
```
currently produces:
```python
([({u'expect': u'a', u'grade_decimal': 1, u'msg': u'', u'ok': True},),
  ({u'expect': u'b', u'grade_decimal': 1, u'msg': u'', u'ok': True},)],)
```
whereas in the overhaul, it produces:
```python
({u'expect': [({u'expect': u'a',
                u'grade_decimal': 1,
                u'msg': u'',
                u'ok': True},),
              ({u'expect': u'b',
                u'grade_decimal': 1,
                u'msg': u'',
                u'ok': True},)],
  u'grade_decimal': 1,
  u'msg': u'',
  u'ok': True},)
```
Hence, we gain the ability to apply grade_decimals and messages at higher levels, so something like
```python
grader = SingleListGrader(
    answers={'expect':['a', 'b'], 'msg':'Ok!},
    subgrader=StringGrader()
)
```
is accepted.

### DIFF
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -467,6 +467,13 @@ class ItemGrader(AbstractGrader):
             answer that received a grade of 0 (default "")
     """
 
+    def __init__(self, config=None, **kwargs):
+        """
+        Validate the ItemGrader's configuration, then call post-schema validation.
+        """
+        super(ItemGrader, self).__init__(config, **kwargs)
+        self.config['answers'] = self.post_schema_validation(self.config['answers'])
+
     @property
     def schema_config(self):
         """
@@ -494,6 +501,21 @@ class ItemGrader(AbstractGrader):
             answer_tuple = (answer_tuple,)
         schema = Schema((self.validate_single_answer,))
         return schema(answer_tuple)
+
+    def post_schema_validation(self, answer_tuple):
+        """
+        Perform any post-schema validation on the answer tuple. This function should be
+        idempotent, as it may be called on a configuration multiple times.
+
+        Any post-schema validation that doesn't work with the answer tuple should be
+        handled in a shadowed __init__ function. The reason this function is invoked
+        separately is so that it can also be called on inferred answers.
+
+        Returns the validated answer_tuple.
+
+        This function exists to be shadowed.
+        """
+        return answer_tuple
 
     def validate_single_answer(self, answer):
         """
@@ -675,6 +697,9 @@ class ItemGrader(AbstractGrader):
             self.config['answers'] = self.schema_answers(inferred)
             # Note that this answer is now stored for future calls, but
             # will be overridden if a new expect value is provided.
+
+            # Perform post-schema validation
+            self.config['answers'] = self.post_schema_validation(self.config['answers'])
 
             # Mark that we are using inferred answers
             self.inferring_answers = True
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -289,11 +289,11 @@ class AbstractGrader(ObjectWithSchema):
         if 'input_list' in result:
             # Multiple inputs
             for idx, entry in enumerate(result['input_list']):
-                cleaned = {key: val for key, val in entry.iteritems() if key in keys}
+                cleaned = {key: val for key, val in entry.items() if key in keys}
                 result['input_list'][idx] = cleaned
         else:
             # Single input
-            result = {key: val for key, val in result.iteritems() if key in keys}
+            result = {key: val for key, val in result.items() if key in keys}
 
         # Handle partial credit based on attempt number
         if self.config['attempt_based_credit']:
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -283,6 +283,18 @@ class AbstractGrader(ObjectWithSchema):
                     formatted = msg.format(student_input)
                 raise StudentFacingError(formatted)
 
+        # Make sure we're only returning the relevant keys in the result.
+        # List graders may use other keys to track information between nesting levels.
+        keys = ['ok', 'grade_decimal', 'msg']
+        if 'input_list' in result:
+            # Multiple inputs
+            for idx, entry in enumerate(result['input_list']):
+                cleaned = {key: val for key, val in entry.iteritems() if key in keys}
+                result['input_list'][idx] = cleaned
+        else:
+            # Single input
+            result = {key: val for key, val in result.iteritems() if key in keys}
+
         # Handle partial credit based on attempt number
         if self.config['attempt_based_credit']:
             self.apply_attempt_based_credit(result, kwargs.get('attempt'))
@@ -472,7 +484,7 @@ class ItemGrader(AbstractGrader):
         Validate the ItemGrader's configuration, then call post-schema validation.
         """
         super(ItemGrader, self).__init__(config, **kwargs)
-        self.config['answers'] = self.post_schema_validation(self.config['answers'])
+        self.config['answers'] = self.post_schema_ans_val(self.config['answers'])
 
     @property
     def schema_config(self):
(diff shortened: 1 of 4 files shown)


jolyonb:
This is now working! It was much easier, cleaner and quicker than I had anticipated!

I still need to write tests...

jolyonb:
@ChristopherChudzicki Ready for review. I think you'll like it. For a major overhaul, this is a whole lot simpler than you may expect! (pun intended ;-)

mitxgraders/listgrader.py
```diff
@@ -758,9 +756,18 @@ def check_response(self, answer, student_input, **kwargs):
...
-        result = consolidate_cfn_return(input_list,
-                                        n_expect=len(answers),
-                                        partial_credit=self.config['partial_credit'])
+        # Consolidate the separate results into a single result
+        result = consolidate_single_return(input_list,
+                                           n_expect=len(answers),
+                                           partial_credit=self.config['partial_credit'])
+
+        # If the result is ok, display the message for this answer
+        if result['ok'] is True and msg != '':
```
jolyonb: This is the only thing I'm in two minds about. I'd like the message to display if the student responses matched their expect values, and I'm using 'ok' as a proxy for this. However, if grade_decimal for a response is less than 1, then I won't catch it. There's just no way of reporting that the entire answer matched something, irregardless of the grade_decimal that was assigned to it... Any thoughts?

jolyonb:
@ChristopherChudzicki Just wondering if you plan to review this, or should I leave it sit a few days and then self-review?

mitxgraders/listgrader.py
jolyonb: Something similar to this. You want `nice job!` to display for `cat, dog` and `cat, DOG`, even though `dog` doesn't quite get full credit. You can't associate a message with `dog` or `DOG` directly, because then `unicorn, dog` and `unicorn, DOG` will both get that message.

ChristopherChudzicki: In the  situation above, if I wanted to give a message on “cat, DOG” but not “unicorn, DOG” then I  would use a tuple of answers in the outermost SingleListGrader. 

Does that make sense? Sorry I can’t give a code example now, on my phone

ChristopherChudzicki:
I’ll try to review it tonight… I started reading through it this morning… But didn’t finish. If I don’t review it by tomorrow morning then I probably won’t for at least a week, so happy to have you do it.

But, I am reasonably confident I can do it tonight

mitxgraders/listgrader.py
jolyonb: Here's what I have in mind exactly:
```python
grader = SingleListGrader(
    answers=(
        {
            'expect': [('X', {'expect': 'x', grade_decimal:0.5}),
                       ('Y', {'expect': 'y', grade_decimal:0.5})], 
            'msg': 'Good pairing!'
        },
        {
            'expect': [('W', {'expect': 'w', grade_decimal:0.5}),
                       ('Z', {'expect': 'z', grade_decimal:0.5})], 
            'msg': 'Good pairing!'
        }
    ),
    subgrader=StringGrader()
)
```
The message should appear for `X, Y`, `x, Y`, `X, y`, and `x, y`, or similarly for `W/w` and `Z/z`. If you get the pairing wrong though, say, `X, Z`, you don't get a message.

Currently, the message will only appear for `X, Y` or `W,Z`, because those are the only combination that awards full credit.

Happy to wait for you to take a look tonight - go and enjoy your day!

ChristopherChudzicki: Ah, I understand. Well, to your original question:

> This is the only thing I'm in two minds about. I'd like the message to display if the student responses matched their expect values, and I'm using 'ok' as a proxy for this. However, if grade_decimal for a response is less than 1, then I won't catch it. There's just no way of reporting that the entire answer matched something, irregardless of the grade_decimal that was assigned to it... Any thoughts?

I agree that would be a more logical way to show messages.  I think that `input_list` (before consolidation) contains all the information you need. Something like

```python
if all(r['grade_decimal']>0 for r in input_list) and msg != '':
    result['msg'] = msg if result['msg'] == '' else result['msg'] + '\n' + msg
```
(or check that r['ok'] is truthy should be fine, too)

mitxgraders/listgrader.py
```diff
@@ -633,37 +640,29 @@ def schema_config(self):
...
             Required('partial_credit', default=True): bool,
-            Required('subgrader'): ItemGrader,
-            Required('answers', default=[]): Any(list, (list,))  # Allow for a tuple of lists
+            Required('subgrader'): ItemGrader
         })
+        # Note that we use the ItemGrader definitions for 'answers', although
+        # validate_expect is shadowed, and we do post-processing in post_schema_validation
 
     # Make sure that the ItemGrader validate_expect isn't used
     validate_expect = None
```
ChristopherChudzicki: This `validate_expect = None` should be deleted now, right?

mitxgraders/baseclasses.py
```diff
@@ -495,6 +502,21 @@ def schema_answers(self, answer_tuple):
         schema = Schema((self.validate_single_answer,))
         return schema(answer_tuple)
 
+    def post_schema_validation(self, answer_tuple):
```
ChristopherChudzicki: Is `post_schema_answer_validation` too long a name? Some graders do post-schema validation that is not answer validation, so it is "post schema validation" but it is not covered by this function.

mitxgraders/listgrader.py
jolyonb: I considered this, but I think it fails for nested `SingleListGrader`s, where partial credit might just mean that you got one wrong?

mitxgraders/baseclasses.py
jolyonb: Indeed, this is only post schema answer validation (only required because of the possibility of inferring answers). I thought about your name, but it was already leading to overly long lines! Maybe I'll call it `post_schema_ans_val`, which isn't quite so long, but still conveys the relevant information.

mitxgraders/listgrader.py
jolyonb: Absolutely. Didn't even realize that was there.

jolyonb:
Ok, I fixed those two little things. I basically did your suggestion for checking whether or not to add the message, but I had to complicate it somewhat in order to account for nested SingleListGraders... Basically, I added an extra field to the return dictionary that recorded the required information, and then stripped it out in the __call__ method.

jolyonb:
@ChristopherChudzicki Updated and ready for final review.

mitxgraders/listgrader.py
```diff
@@ -761,9 +758,25 @@ def check_response(self, answer, student_input, **kwargs):
...
                                            partial_credit=self.config['partial_credit'])
 
-        # If the result is ok, display the message for this answer
-        if result['ok'] is True and msg != '':
-            result['msg'] = msg if result['msg'] == '' else result['msg'] + '\n' + msg
+        # Check if all inputs were awarded credit
+        all_awarded = False
+        if not isinstance(self.config['subgrader'], SingleListGrader):
+            # Check to see if all items were awarded credit
+            if all(item['grade_decimal'] > 0 for item in input_list):
```
ChristopherChudzicki: Minor comment: here and below could just do 

`all_awarded=all(item['grade_decimal'] > 0 for item in input_list)`

I believe no need for conditional. Also no need for initialization to False above.

jolyonb: Good point - fixed.

ChristopherChudzicki:
Nice resolution to the message issue!

PR #269, opened 2019-07-12
Fixing bug with LinearComparer with less than 3 samples

jolyonb:
Caught a little bug in LinearComparer when writing documentation. We need to have at least 3 samples for the comparison to be meaningful!

In particular, NumericalGrader/MatrixGrader/FormulaGrader all shared the same default grader, if it was set on FormulaGrader. This decouples everything.


ChristopherChudzicki:
Good idea!

PR #273, opened 2019-07-15
DependentSampler upgrades

### LINKED ISSUE #271: Upgrading DependentSampler
Two things we can do to improve `DependentSampler`:

- Allow user functions to be used in `DependentSampler`
- Automatically detect the `depends` list

jolyonb:
This does two things:

- Makes DependentSampler parse the formula to compute dependencies
- Gives DependentSampler access to user_functions (this was actually already done, but wasn't documented properly!)

Also provides some minor linting.

Resolves #271


ChristopherChudzicki:
Nice improvement!

PR #277, opened 2019-07-18
Coerce all inputs to unicode

### LINKED ISSUE #238: Fix unicode issue with schemas
@ChristopherChudzicki  pointed out an issue with `Any(None, 'upper', 'lower')` as a schema in #222 with regards to unicode. I'm not sure what it is, but here's a ticket for it anyway.

jolyonb:
This coerces all arguments provided in `ObjectWithSchema` configs to unicode as necessary.

I don't know a good way to test that this works explicitly. Suggestions welcome.

Resolves #238

### DIFF
--- a/mitxgraders/baseclasses.py
+++ b/mitxgraders/baseclasses.py
@@ -18,6 +18,7 @@ from voluptuous.humanize import validate_with_humanized_errors as voluptuous_val
 from mitxgraders.version import __version__
 from mitxgraders.exceptions import ConfigError, MITxError, StudentFacingError
 from mitxgraders.helpers.validatorfuncs import text_string, Positive
+from mitxgraders.helpers.compatibility import ensure_text
 
 class DefaultValuesMeta(abc.ABCMeta):
     """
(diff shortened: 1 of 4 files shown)


ChristopherChudzicki:
This looks good to me. Probably could have done this from the beginning rather than replacing `str` with `text_string`. Oh well.

If you want to test this explicitly, I think you'd need to do something like:

```python
import six
from voluptuous import Schema, Required
from mitxgraders.baseclasses import ObjectWithSchema

def test_coerce2unicode():
    class Foo(ObjectWithSchema):
        schema_config = Schema({
            Required('a'): 'cat',
            Required('b'): ['bear', 'whale'],
            Required('c'): ['otter', 'seal'],
            Required('d'): {'moose': 'antlers'}
        })

    foo = Foo(
        a='cat',
        b=['bear', 'whale'],
        c=['otter', 'seal'],
        d={'moose': 'antlers'}
    )

    config = foo.config
    key_unicode = all([isinstance(x, six.text_type) for x in config])
    assert sum(key_unicode) == 4
    
    assert isinstance(foo.config['a'], six.text_type)
    
    b_unicode = all([isinstance(x, six.text_type) for x in config['b']])
    assert sum(b_unicode) == 2

    # etc
```

I'm going to merge this, but if you want to add something like that in another PR yourself and merge it, go for it.

PR #278, opened 2019-07-18
Adding extra error check to DependentSampler, fixing test coverage

jolyonb:
This adds an extra check to DependentSampler that we somehow missed previously. Also updates tests; should get us back to 100% coverage.


ChristopherChudzicki:
Looks good 👍

PR #280, opened 2019-07-20
Generalizing attempt-based partial credit

jolyonb:
This overhauls attempt-based partial credit. I think it's an improvement over the API I previously implemented, as it allows for maximum flexibility, while useful defaults are provided.


jolyonb:
So, apparently rounding is slightly different between python 2 and 3. Fixed tests!

jolyonb:
The coding here is all straightforward; most of this PR is documentation and examples. I'm going to unilaterally merge it :-)

Issue #283, opened 2019-07-25
Preprocessor wrapping and sqrt


jolyonb:
We shouldn't do any wrapping of the sqrt function, as `{:sqrt:}` gives an error (even in raw asciimath).

Also, `x/sqrt(` gives an asciimath error in our script on edX. I'm not sure what the processed output looks like though; I thought that `sqrt(` wouldn't get any wrapping, and I can't seem to reproduce this one in raaw asciimath.

jolyonb:
Resolved by #284

PR #284, opened 2019-08-01
do not wrap lonesome sqrts

ChristopherChudzicki:
Addresses #283 

I'm surprised we didn't catch this before 🤷‍♂ 

I tested this on my personal version of demo course, but have not pushed to "public" demo course.

I also bumped version number and changelog. (And I found 3 typos. So I've found 3, and you've found 500 of mine 😄 )

### DIFF
--- a/MJxPrep.js
+++ b/MJxPrep.js
@@ -292,7 +292,7 @@ if (window.MJxPrep) {
   var checkExist = setInterval(updateMathJax, 200);
 
   // FORBIDDEN_TO_WRAP includes things we've found problematic to wrap {:...:},
-  // namelu:
+  // namely:
   // - AsciiMath symbols classified as mover (math-overscript) tags
   // - sqrt symbol (msqrt tag)
   // (This should be a Set object, but ie11 doesn't fully support Sets)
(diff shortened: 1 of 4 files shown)


MJxPrep.js
```diff
@@ -291,9 +291,12 @@ if (window.MJxPrep) {
   // Check for the AsciiMath object every 200ms
   var checkExist = setInterval(updateMathJax, 200);
 
-  // set of AsciiMath symbols classified as mover (math-overscript) tags
+  // FORBIDDEN_TO_WRAP includes things we've found problematic to wrap {:...:},
+  // namelu:
```
jolyonb: namely?

MJxPrep.js
```diff
@@ -394,7 +398,8 @@ if (window.MJxPrep) {
 
       var front = eqn.slice(0, funcStart)
       var middle = eqn.slice(funcStart, funcEnd)
-      eqn = front + wrapIfNotOverscript(null, middle) + eqn.slice(funcEnd)
+      console.log(middle)
```
jolyonb: Probably don't want this left here?

MJxPrep.js
jolyonb: @ChristopherChudzicki Bump?

MJxPrep.js
jolyonb: @ChristopherChudzicki Bump?

MJxPrep.js
ChristopherChudzicki: fixed, sorry slow

MJxPrep.js
ChristopherChudzicki: fixed

PR #285, opened 2019-08-23
Allow SingleListGrader answers to be specified as a string

jolyonb:
This is useful when providing grade_decimal or msg entries in answers, as it means you don't need to construct a list of individual entries, but rather just specify the string that is expected from students.


ChristopherChudzicki:
👍

PR #286, opened 2019-08-23
IntervalGrader

jolyonb:
This implements IntervalGrader. It's based on a SingleListGrader. Students type in `[1, 2)` or similar. By default, comparisons are done with high (absolute) precision NumericalGraders, but FormulaGraders can also be used. I made the decision that the bracket and the bound are treated as one unit for the purpose of partial credit.

Tests included, but no documentation/examples yet. That'll be next.

### DIFF
--- a/mitxgraders/formulagrader/intervalgrader.py
+++ b/mitxgraders/formulagrader/intervalgrader.py
@@ -13,6 +13,7 @@ from mitxgraders.stringgrader import StringGrader
 from mitxgraders.listgrader import SingleListGrader, demand_no_empty
 from mitxgraders.formulagrader.formulagrader import FormulaGrader, NumericalGrader
 from mitxgraders.exceptions import ConfigError, InvalidInput, MissingInput
+from mitxgraders.helpers.validatorfuncs import text_string
 
 class IntervalGrader(SingleListGrader):
     """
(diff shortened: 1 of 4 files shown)


mitxgraders/formulagrader/intervalgrader.py
```diff
@@ -0,0 +1,231 @@
...
+
+class IntervalGrader(SingleListGrader):
+    """
+    IntervalGrader grades a pair of numbers describing an interval.
+    Both the interval bracketing and the numbers are graded.
+    The following are examples of possible entries:
+    * [1, 2)
+    * {0, 1+pi]
+
+    Answers should be provided as a list of four entries:
```
ChristopherChudzicki: I see that `IntervalGrader` subclasses `SingleListGrader`. Still, this seemed to me like an unexpected way to specify answers when grading an interval.

Thought: Is it worth having answers accept `[a, b)`, and just coercing to this form with a `schema_expect` if this form is needed internally?

(Would save from documenting two separate forms of the answer, too.)

mitxgraders/formulagrader/intervalgrader.py
```diff
@@ -0,0 +1,231 @@
...
+from mitxgraders.exceptions import ConfigError, InvalidInput, MissingInput
+from mitxgraders.helpers.validatorfuncs import text_string
+
+class IntervalGrader(SingleListGrader):
+    """
+    IntervalGrader grades a pair of numbers describing an interval.
+    Both the interval bracketing and the numbers are graded.
+    The following are examples of possible entries:
+    * [1, 2)
+    * {0, 1+pi]
```
ChristopherChudzicki: The `{` isn't a typo, is it? You're pointing out that `{` is possible with configuration?

(Incidentally, `{` seems useful for discrete sets, but in that scenario we wouldn't want the two-element limitation. I suppose that is out-of-scope for the current work, though)

tests/formulagrader/test_intervalgrader.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+from __future__ import print_function, division, absolute_import
+
+from pytest import raises
+from voluptuous import Error
+
+from mitxgraders import IntervalGrader, NumericalGrader, FormulaGrader, StringGrader
+from mitxgraders.exceptions import ConfigError, InvalidInput
+
+def test_usage():
```
ChristopherChudzicki: I totally believe that it works, but do you want to add a test using FormulaGrader with variables?

mitxgraders/formulagrader/intervalgrader.py
jolyonb: There's a reason to have all four entries separated in a list (or at least, the ability to do so) - so you can give partial credit/messages on incorrect brackets etc (the full ItemGrader format is available to each entry).

Regardless, now that the other PR is merged, I've added the other way to specify the answer: as a string that you would expect from students.

mitxgraders/formulagrader/intervalgrader.py
jolyonb: I'll change it to a default setting, and add that configuration options will allow other brackets (eg, `{}` and `<>`).

tests/formulagrader/test_intervalgrader.py
jolyonb: Done.

jolyonb:
I've pushed my updates for the moment. I'm going to go and play with infinities now...

jolyonb:
Ok, I went and implemented #287 to handle infinities properly. Once that's merged, I'll update the defaults here to allow for infinities 😄

jolyonb:
Ok, you were too fast on merging this. I just pushed directly to master to get the defaults to allow infinities. I appreciate you taking the time to check my work!

ChristopherChudzicki:
Ah sorry, I was wondering about default inf ... sounds good :)

PR #293, opened 2020-03-26
Implementing sumgrader

jolyonb:
What it says on the tin. Includes tests, but no documentation/examples as of yet.

### DIFF
--- a/mitxgraders/formulagrader/integralgrader.py
+++ b/mitxgraders/formulagrader/integralgrader.py
@@ -11,20 +11,21 @@ from functools import wraps
 from scipy import integrate
 from numpy import real, imag
 from abc import abstractproperty
+from numbers import Number
 
 from voluptuous import Required, Any, Extra
 
 from mitxgraders.baseclasses import AbstractGrader
 from mitxgraders.comparers import equality_comparer
 from mitxgraders.exceptions import (InvalidInput, ConfigError,
-                                    StudentFacingError, MissingInput)
-from mitxgraders.helpers.validatorfuncs import Positive, text_string
+                                    StudentFacingError, MissingInput, MITxError)
+from mitxgraders.helpers.validatorfuncs import Positive, text_string, NonNegative, PercentageString
 from mitxgraders.helpers.math_helpers import MathMixin
 from mitxgraders.helpers.calc import evaluator, DEFAULT_VARIABLES
 from mitxgraders.helpers.calc.mathfuncs import merge_dicts
 
 
-__all__ = ["IntegralGrader"]
+__all__ = ['IntegralGrader', 'SumGrader']
 
 def is_valid_variable_name(varname):
     """
@@ -73,6 +74,10 @@ def transform_list_to_dict(thelist, thedefaults, key_to_index_map):
 
 class IntegrationError(StudentFacingError):
     """Represents an error associated with integration"""
+    pass
+
+class SummationError(StudentFacingError):
+    """Represents an error associated with summation"""
     pass
 
 def check_output_is_real(func, error, message):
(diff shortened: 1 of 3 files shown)


jolyonb:
This isn't as bad as it might look on first blush; the vast majority of the commit is the testing suite (a lot of which was copied from IntegralGrader).

mitxgraders/formulagrader/integralgrader.py
```diff
@@ -149,6 +154,15 @@ def get_limits_and_funcs(self, integrand_str, lower_str, upper_str, dummy_var,
...
         varscope[dummy_var] = (upper + lower) / 2
+        if varscope[dummy_var] == float('inf') or varscope[dummy_var] == -float('inf'):
+            if -float('inf') < lower < float('inf'):
+                varscope[dummy_var] = lower
+            elif -float('inf') < upper < float('inf'):
+                varscope[dummy_var] = upper
+            else:
+                # Try this instead?
+                varscope[dummy_var] = 0
+
```
ChristopherChudzicki: **Requested Change:** Initially, I found this a bit confusing ... why do you need the dummy variable value outside of `eval_sum` / `eval_integrand`. I believe you're doing this to get at the `integrand_used` variables below.

An alternative would be to use `PARSER.parse(formula).variables_used` to get the variables used without actually evaluating. Although, I do not love importing/exporting our global parser instance all over the place. So I just made a PR that exports a `parse` function from expressions.py. See https://github.com/mitodl/mitx-grading-library/pull/294

So this could be:
```python
integrand_used = parse(integrand_str).variables_used
```

mitxgraders/formulagrader/integralgrader.py
```diff
@@ -537,3 +551,326 @@ def raw_integrand(x):
...
+                                 functions=funcscope,
+                                 suffixes=self.suffixes)
+            del varscope[summation_var]
+            return value
+
+        # Check if used_funcs includes a factorial function
+        if 'fact' in used_funcs or 'factorial' in used_funcs:
+            infty_val = self.config['infty_val_fact']
+        else:
+            infty_val = self.config['infty_val']
```
ChristopherChudzicki: **Question:** Is `infty_val_fact` really worthwhile? Authors could just set `infty_val` to be small themselves. My main "concern" would just be that this gives a false sense of security.... there are super-factorial terms that are at least somewhat common. E.g., sums in calculus textbooks often have terms like `n^n`.

But... I am not particularly familiar with the actual problems people will be writing with this. So maybe it's totally reasonable.

mitxgraders/formulagrader/integralgrader.py
ChristopherChudzicki: I just saw:

> I'm guessing this was annoying you...

Yes 😛

mitxgraders/formulagrader/integralgrader.py
```diff
@@ -537,3 +551,326 @@ def raw_integrand(x):
...
+    numerical error to creep in. Make sure that your tolerance is sufficiently
+    large to satisfy such situations. When in doubt, use debug mode to view
+    what the sums are actually evaluating to.
+    
+    Configuration Options
+    =====================
+        answers (dict, required): Specifies author's answer. Has required keys lower,
+            upper, summand, summation_variable, which each take string values.
+
+        input_positions (dict): Specifies which integration parameters the student
```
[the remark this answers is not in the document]
jolyonb: Good catch!

mitxgraders/formulagrader/integralgrader.py
```diff
@@ -537,3 +551,326 @@ def raw_integrand(x):
...
+        even_odd (int): Choose to sum every number (0), every odd number (1) or every even number (2).
+
+    Additional Configuration Options
+    ================================
+    The configuration keys below are the same as used by FormulaGrader and
+    have the same defaults, except where specified
+        user_constants: same as FormulaGrader, but with additional default 'infty'
+        whitelist
+        blacklist
+        tolerance (default changed to 1e-12)
```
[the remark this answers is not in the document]
jolyonb: This is in place because of the way that infinite sums are treated. If an infinite sum converges, then the terms should be getting smaller and smaller after some point. Because of the way that infinite sums can be rewritten, the comparison between the student and instructor answers may include different numbers of terms. The idea of an absolute tolerance is to allow for the different number of terms.

tl,dr; infinite sums are tricksy things and I think an absolute tolerance works better there.

mitxgraders/formulagrader/integralgrader.py
jolyonb: The idea is to try to detect when an overflow/underflow is going to occur in an infinite summation, and use a smaller number to prevent it from happening. I'd prefer that students not get overflow messages that are confusing. The most common source of that will be from factorials, I suspect, which saves the problem author from having to worry about setting `infty_val` each time. I'm not sure that superfactorial type terms will show up all that often in an infinite summation, no? Whereas factorials are very, very common in infinite summations.

mitxgraders/formulagrader/integralgrader.py
jolyonb: This is a significantly better approach - no need to evaluate the expression here at all, when all we want is the list of functions used. Implementing now!

jolyonb:
Fixed the things you pointed out and rebased on top of your other PR.

mitxgraders/formulagrader/integralgrader.py
ChristopherChudzicki: you're probably right that this will show up much more often, and that makes sense.

Superfactorial terms like `n^n` show up a lot in calculus books when doing convergence tests (i.e., does it converge, not "what does it convert to"). But I dunno if they'll show up in SumGrader problems.

PR #304, opened 2020-11-15
Input sanitizing, parsing emdash as minus sign, adds some tests for that.

drhodes:
Hi!  Someone reported an issue when pasting some math into an input field on edx.  Some confusion arose around a unicode look-a-like character that looks like a dash, but is, in fact an emdash!  Here they are:

`emdash: –` 
`dash..:      -`

The problem was that the mitx-grading-library parser was expecting an ascii dash but instead got an emdash.  So, the user got a syntax error.  This commit fixes that problem by adding a rule that tokenizes emdashes and immediately converts them to ascii dashes.

### DIFF
--- a/mitxgraders/helpers/calc/expressions.py
+++ b/mitxgraders/helpers/calc/expressions.py
@@ -387,7 +387,12 @@ class MathParser(object):
 
         # Define + and -
         plus = Literal("+")
-        minus = Literal("-")
+
+        # Also accept unicode emdash
+        emdash = Literal("\u2014") 
+        emdash.setParseAction(lambda: "-")
+        
+        minus = Literal("-") | emdash
         plus_minus = plus | minus
 
         # 1 or 1.0 or .1
--- a/tests/helpers/calc/test_expressions.py
+++ b/tests/helpers/calc/test_expressions.py
@@ -1,7 +1,7 @@
 """
 Tests of expressions.py that aren't covered elsewhere
 """
-from __future__ import print_function, division, absolute_import
+from __future__ import print_function, division, absolute_import, unicode_literals
 
 import re
 import numpy as np
@@ -147,6 +147,35 @@ def test_negation():
         with raises(UnableToParse, match=re.escape(msg.format(formula))):
             evaluator(formula)
 
+
+def test_negation_with_emdash():
+    """Test that appropriate numbers of +/emdash signs are accepted"""
+    assert evaluator(u"1+\u20141")[0] == 0
+    assert evaluator("1\u2014\u20141")[0] == 2
+    assert evaluator("2*\u20141")[0] == -2
+    assert evaluator("2/\u20144")[0] == -0.5
+    assert evaluator("\u20141+\u20141")[0] == -2
+    assert evaluator("+1+\u20141")[0] == 0
+    assert evaluator("2^\u20142")[0] == 0.25
+    assert evaluator("+\u20141")[0] == -1
+
+    badformulas = [
+        "1\u2014\u2014\u20141",
+        "2^\u2014\u20142",
+        "1\u2014+2",
+        "1++2",
+        "1+++2",
+        "\u2014\u20142",
+        "\u2014\u2014\u20142",
+        "\u2014+2",
+        "\u2014\u2014+2",
+    ]
+
+    msg = "Invalid Input: Could not parse '{}' as a formula"
+    for formula in badformulas:
+        with raises(UnableToParse, match=re.escape(msg.format(formula))):
+            evaluator(formula)
+            
 def test_evaluation_does_not_mutate_variables():
     """
     This test should not be considered as related to vector/matrix/tensor algebra.


jolyonb:
Hi @drhodes! Thanks for submitting our first community contribution!

Tests are presently failing for python 2.7. Here's the error log:

```text
    def test_negation_with_endash():

        """Test that appropriate numbers of +/– signs are accepted"""

>       assert evaluator("1+–1")[0] == 0

tests/helpers/calc/test_expressions.py:153: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

mitxgraders/helpers/calc/expressions.py:1243: in evaluator

    parsed = parse(formula)

mitxgraders/helpers/calc/expressions.py:1166: in parse

    return PARSER.parse(formula)

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <mitxgraders.helpers.calc.expressions.MathParser object at 0x7f4a20f4e990>

expression = '1+\xe2\x80\x931'

    def parse(self, expression):

        """

        If expression is in parser cache, return cached result, otherwise

        delegate to raw_parse.

        """

>       expression_no_whitespace = expression.replace(' ', '')

E       UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 2: ordinal not in range(128)

mitxgraders/helpers/calc/expressions.py:537: UnicodeDecodeError
```

Dealing with unicode simultaneously in python 3 and 2.7 is a pain, but it's something we need to maintain for backwards compatibility. Can I get you to modify things as necessary?

mitxgraders/helpers/calc/expressions.py
```diff
@@ -387,7 +388,12 @@ def get_grammar(self):
 
         # Define + and -
         plus = Literal("+")
-        minus = Literal("-")
+
+        # Also accept unicode emdash
+        emdash = Literal("–")
```
jolyonb, line 393: I would MUCH rather use `"\u2014"` here, rather than setting an encoding on the script.

jolyonb: Because we're importing unicode_literals, this should work in both python 2.7 and 3.

drhodes: Excellent! Hey! I took 8.02x, it was brilliant, thank you.   About the encoding, yes, I'm happy to make those changes. Sorry for making so much noise with the build, I'm still kicking the tires on codecov and Travis CI. 

Enough excuses from me, hopefully the next commit will be all green.  Let me know if you would prefer that I retract this pull request and squash the previous two commits.

jolyonb: Keep the PR - it's got this discussion in it. Once you're all green, we'll squash before merging :-) If you activate a python 2 virtual environment, you can test locally there using pytest.

drhodes: Thanks, I'm all set on testing with python2 now, locally using pyenv with 2.7.15 to match the build bot.

drhodes: There is a small issue outstanding with the new test. I'm not sure how
to fix it. There is a test function found in test_expressions.py
called:

test_negation_with_emdash()

It loops through badformulas testing to make sure badly formatted
formulas are unable to parse.  That bit works, that is if the test
fails then exceptions are raised, however, pytest has a feature that
should be used, but is not in this case. A particular function:
pytest.raises can take a keyword argument called "match" that ensures
error messages raised along with the parse error are formatted
correctly.

There is or was a bug that seems to mirror the issue here:

https://github.com/pytest-dev/pytest/issues/5478

drhodes:
Ok, had some time today and fixed this by abandoning the "pytest.raises" function. Now manually checking the error message and failing the test if the strings don't match.  Something to watch out for, if an error with a unicode emdash bubbles up to the browser - will it render for the user as "\u2014" for python2 people?  I'm starting to question whether or not this was a good idea. Maybe it would be better to just kill this pull request and wait until edx drops python2 support!

jolyonb:
Sorry I'm taking a while to get to this... looking into it now!

jolyonb:
Ok, so the issue with `raises` was a bug in an older version of pytest. I'm updating the version of pytest in another PR. Can I get you to rebase and squash this PR please?

tests/helpers/calc/test_expressions.py
```diff
@@ -1,11 +1,11 @@
 """
 Tests of expressions.py that aren't covered elsewhere
 """
-from __future__ import print_function, division, absolute_import
+from __future__ import print_function, division, absolute_import, unicode_literals
 
 import re
 import numpy as np
-from pytest import raises, approx
+from pytest import raises, approx, fail
```
jolyonb, line 8: Now unnecessary.

tests/helpers/calc/test_expressions.py
```diff
@@ -15,6 +15,11 @@
 )
 from mitxgraders.helpers.calc.math_array import equal_as_arrays, MathArray
 
+# python3 doesn't have the unicode function
```
jolyonb, line 18: This can all be removed.

tests/helpers/calc/test_expressions.py
```diff
@@ -147,6 +152,39 @@ def test_negation():
...
+        "1+++2",
+        "\u2014\u20142",
+        "\u2014\u2014\u20142",
+        "\u2014+2",
+        "\u2014\u2014+2",
+    ]
+
+    msg = "Invalid Input: Could not parse '{}' as a formula"
+    for formula in badformulas:
+        try:
```
jolyonb, line 181: Just use these two lines:
```python
        with raises(UnableToParse, match=re.escape(msg.format(formula))):
            evaluator(formula)
```

jolyonb (changes requested):

drhodes:
Oops. Looks like I messed up rebase/squash. Please bear with me on this :)

jolyonb:
No worries :-) I recommend following the steps in this guide: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

jolyonb (approved):

jolyonb:
Thanks for your contribution! I'll make a new version and zip things up shortly :-)

PR #306, opened 2020-11-21
Implementing functionality to remove default math constants

jolyonb:
This allows an instructor to set a pre-defined constant to `None` to remove it from the list of constants.


jolyonb:
@jefrench As you requested!

PR #308, opened 2021-04-06
Fixing bug with long ints

jolyonb:
Previously, long integers passed in sample_from or user_constants would cause an error. This fixes that error.

Code:
```python
from mitxgraders import *

grader = MatrixGrader(
    answers='x',
    user_constants={'x': 10 ** 19}
)

grader(None, 'x')
```


jolyonb:
@ChristopherChudzicki You might find this one amusing.

Issue #309, opened 2021-05-03
Upgrade numpy?


pomegranited:
Hello!

We're trying to upgrade `numpy` as used in the edx-platform sandbox from `1.16.5` to `1.20.2` to [match edx-platform master](https://github.com/edx/edx-platform/blob/3c0cfd87a37de6f8ad4f7c3c105994e921c954dd/requirements/edx/base.txt#L166). This resolves an issue with another sandbox dependency, `matplotlib`, see https://github.com/edx/edx-platform/pull/27372 .

But this update could cause a similar issue with users of this grader, depending on how the dependencies are installed.

Could you help us out by testing the update on https://github.com/edx/edx-platform/pull/27372 to see if it breaks your grader?

Thank you!

jolyonb:
Closing this issue because the use case was impossible to use.

PR #310, opened 2021-05-03
Upgrading numpy version for python3 only

jolyonb:
Testing compatibility re: Issue #309.


jolyonb:
This is no longer required.

Issue #320, opened 2022-11-02
Import time optimization to prevent formatting error.


drhodes:
Hi, the 18x courses on edx have been experiencing an intermittent HTML format error that prevents learners from viewing entire exercises.  This happens anytime the codejail runs for longer than 3 seconds.  Sometimes the error happens during timed final exams, which can be .. awkward.  Anyways, I've been exploring mitigations. The one I'm currently presenting doesn't require course staff to alter grading scripts, and it doesn't require coordinating with the codejail team to optimize anything on their end.

I did some profiling on the mitx-grading-library and found that scipy was causing a large delay on import. Thankfully, scipy accounts for a vanishing portion of the code base.  Here are some numbers for python3.7.  Profiling was done with and without __pycache__ present, and then also with and without lazy scipy imports.  The times are rough averages.

```
| python version | __pycache__ available? | lazy import scipy? | import time (ms) | grader runtime (ms) |
|----------------+------------------------+--------------------+------------------+---------------------|
| python 3.7.15  | no                     | no                 |             1300 |             1400    |
| python 3.7.15  | yes                    | no                 |              365 |              430    |
| python 3.7.15  | no                     | yes                |              530 |              580    |
| python 3.7.15  | yes                    | yes                |              157 |              220    |

```

As far as I can tell, codejail does not use __pycache__.  By lazily importing scipy, I would expect the grader script runtime (at least for this one example) would drop from 1400ms down to about 580ms on edx and would be one big step towards preventing the  HTML format error.  I've made the optimization, for anyone interested the [commit can be found here](https://github.com/drhodes/mitx-grading-library/commit/55ee1abb1d537815bb6165884689ee301204267e).  It's a small change and it doesn't break any tests with python 3.7.15.  I haven't tested with 2.7 because honestly, I can't tell if 2.7 is still used anywhere ..  and when I did try pip installing the 2.7 requirements in a venv, numpy was complaining.  So, I figured I'd get some feedback first.

thoughts?

jolyonb:
Nice work! I agree that loading scipy can be skipped for the vast majority of use cases. The cases you addressed were the three that I thought of - integration, factorial and some fancy matrix sampling. I'm happy to merge this into the publication version. Please make a PR against the base repo and we'll take it from there :)

drhodes:
Hi Jolyon, thanks for the prompt review!

jolyonb:
Addressed in #321.

PR #321, opened 2022-11-03
lazily import scipy to avoid import delay

drhodes:
Addresses #320

### DIFF
--- a/mitxgraders/formulagrader/integralgrader.py
+++ b/mitxgraders/formulagrader/integralgrader.py
@@ -8,7 +8,6 @@ from __future__ import print_function, division, absolute_import, unicode_litera
 
 import six
 from functools import wraps
-from scipy import integrate
 from numpy import real, imag
 from abc import abstractproperty
 from numbers import Number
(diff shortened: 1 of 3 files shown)


jolyonb:
I did check how codejail interacts with scipy to make sure that scipy isn't being imported anyway; it's not.

jolyonb:
Aww man, looks like codecov and travis have both broken :( Guess it's time to update the CV/CD pipeline!

jolyonb:
@drhodes You'll be happy to know that all tests went through successfully on all versions of python :)

jolyonb:
@drhodes I've created a new version/release/zipfile with this update. It should be ready to use immediately.

drhodes:
@jolyonb, thank you so much for taking the time out to do this last night, I really appreciate it!

> I did check how codejail interacts with scipy to make sure that scipy isn't being imported anyway; it's not.

Good thinking on that! How unfortunate that would have been.

> ... codecov and travis have both broken :( Guess it's time to update the CV/CD pipeline!

I don't know anything about that stuff, but I really should. If you would like a hand with that, then please let me know. I'd like to start integrating automated checks with some other projects.

Thanks again!

Issue #326, opened 2024-06-28
Upgrade code and dependencies for Redwood and later releases


blarghmatey:
### Description/Context

The Redwood release updated the version of Python being used in codejail and edx-platfrom from 3.8 to 3.11. Along with that there were version updates to the dependencies installed in the codejail environment, which this library relies on. The current code in this repository is incompatible with the newer versions of Python and dependencies required for Redwood and later. This affects the ability to use this module in courses running on MITx Online, Residential MITx, and will impact the ability to run those courses on edx.org (if not now, then in the near future).

### Plan/Design

- [ ] Tag the current code in the repository to provide a stable target for environments running Quince and older
- [ ] Test the code in this repository in a Python 3.11 environment using the dependency versions specified in https://github.com/openedx/edx-platform/blob/master/requirements/edx-sandbox/releases/redwood.txt
- [ ] Address all exceptions and errors that result from running in that environment
- [ ] Tag the repository with `redwood` to give a stable target for Redwood releases

MaferMazu:
@blarghmatey, there is a release that supports new releases of Open edX. Should we close this issue?

PR #327, opened 2024-11-07
chore: update actions and test with python 311

MaferMazu:
## Description

This PR adds support to Python 3.8 and 3.11, drops support for the previous version of Python, and fixes the issues related to this support.

Related to #326 

### Changes

- Update the workflows.
- Remove the constraints in the requirements.txt to avoid maintaining unnecessary constraints.
- Change `np.complex` to use `complex` based on a deprecation in numpy 1.20. Ref: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
- Replace `imp` with `importlib` because imp was deprecated in Python 3.4. Ref: https://docs.python.org/3/whatsnew/3.4.html#importlib
- Use approx as the doc suggests. Ref: https://docs.pytest.org/en/4.6.x/reference.html#pytest-approx
- Add `pygments` for the doc, and add `vs-min` as the code theme to maintain the doc `pycon` highlight
- Change the `.asXML()` function because it was deprecated and use `xml` and a custom function. Ref: https://github.com/pyparsing/pyparsing/blob/96d44647450463a7b9c0962b03f3aeb0d1f165d5/CHANGES#L1183
    - First, I tried with `dicttoxml`, which already has the function, but when using the `python_lib.zip` file, Studio doesn't encounter that requirement, and we should add it in the openedx sandbox requirements or add it as a codejail requirement and build the image. I propose using the XML package and creating a function to avoid that path.

### Code

- All the checks from Github are passing.

### Integration

**Environment**
- Have a tutor environment with Tutor v18.1.4
- Install codejail. You can follow the installation and usage process from the [tutor-contrib-codejail README](https://github.com/eduNEXT/tutor-contrib-codejail?tab=readme-ov-file#codejail-plugin-for-tutor).
- Perform a dev launch

**Data**
- Import the MIT demo course running:
```
tutor dev do importdemocourse -r https://github.com/mitodl/mitx-grading-library -d course -v master
```
For more info: `tutor dev do importdemocourse -h`
- Create an admin user
```
tutor dev do createuser --staff --superuser admin admin@example.com
```

**Test**
1. Enter in **Studio > MITx Grading Library > Content > Files**
2. Remove the `python_lib.zip` and add this one:
[python_lib.zip](https://github.com/user-attachments/files/17670396/python_lib.zip)
> [!NOTE]  
> If you want to create your python_lib from this version, you can clone the repo, enter in it, and then:
> ```
> find mitxgraders voluptuous -type f ! -path "*/__pycache__/*" | zip python_lib.zip -@
> ```
3. Make sure the units are rendered and worked well.

### Doc

- Create a Python Virtual Environment
- Clone this repo in this branch and enter with `cd`.
- Run `pip install -r requirements.txt`.
- Run `mkdocs serve`.
- Compare the doc with the actual one: https://mitodl.github.io/mitx-grading-library/

## Screencasts

I test the StringGraders, MatrixGraders, and SumGraders in the following video:

[Screencast from 2024-11-07 18-23-57.webm](https://github.com/user-attachments/assets/4003bddc-6c89-4018-b5bb-f520bc2e7481)

## Things checked

- [x] Update basic doc
- [x] Test doc in local
- [x] Test in an environment

## Following work

- Create a PR to remove the Python 2 exceptions and considerations from the code.
- Bump version to 3.0.0 because we have breaking changes and edited the doc.


jolyonb:
Oof, lots of work here! I completely understand why you called this a chore!

Will try to take a detailed look through this over the weekend.

README.md
```diff
@@ -6,7 +6,7 @@ A library of graders for edX Custom Response problems.
 
 Version 2.4.1 ([changelog](docs/changelog.md))
 
-Copyright 2017-2022 Jolyon Bloomfield and Chris Chudzicki
+Copyright 2024
```
jolyonb, line 9: Aww man, did you need to remove me? ;-)

mitxgraders/helpers/calc/expressions.py
```diff
@@ -586,7 +590,22 @@ def __init__(self, expression, tree, variables_used, functions_used, suffixes_us
         self.tree = tree
 
     def __str__(self):
-        return self.tree.asXML()
+        tree_dict = self.tree.as_dict()
```
jolyonb, line 593: To be quite honest, these functions were here for debugging purposes only, and could be removed completely. I'd actually suggest commenting them out so that the xml library doesn't need to be loaded, because the grader is on a short timeout, and loading that library isn't worth it.

requirements.txt
jolyonb, line 1: I would STRONGLY recommend having separate versions of `requirements.txt` for python 3.8 and python 3.11, with library dependencies pinned to those explicitly used in the edX versions that use those python versions.

requirements.txt
```diff
@@ -1,19 +1,17 @@
...
-pyparsing==2.0.7
-scipy==0.14.0
-six==1.11.0
+# Used by the library
+numpy
+scipy
+pyparsing
+six
 
 # For testing
```
jolyonb, line 7: Testing and documentation can be unpinned.

Was mock moved into the standard library?

jolyonb:
Okay, this wasn't as big as I suspected it might be. Only minor fixes required at this stage.

> Create a PR to remove the Python 2 exceptions and considerations from the code.

This one will be big though. I will be SO happy to see the `six` package excised from this library!!!

> Bump version to 3.0.0 because we have breaking changes and edited the doc.

Look forward to seeing this happen too :)

README.md
MaferMazu: Sorry for that, I fixed it 🙈

mitxgraders/helpers/calc/expressions.py
MaferMazu: So, we want to comment the `str` method and edit this doc test, right? To be able to avoid the import of xml librery.

Test: https://github.com/mitodl/mitx-grading-library/blob/065bc457f622a38fb8cea14449bdddb439be12ca/mitxgraders/helpers/calc/expressions.py#L312

MaferMazu: I'll comment it.

requirements.txt
MaferMazu: Yes, mock comes from https://docs.python.org/3/library/unittest.mock.html.

MaferMazu:
I applied the feedback. I would like a re-review @jolyonb

README.md
```diff
@@ -31,15 +31,15 @@ A demonstration course for the MITx Grading Library can be viewed [here](https:/
...
 To install:
 
-**Requirements:** An installation of Python 2.7 (for older edX versions) or Python 3.6 or 3.7 (current edX versions).
+**Requirements:** An installation of Python 3.8 or 3.11 (current edX versions).
 
-0. (Optional) Create and activate a new python virtual environment.
+0. (Optional) Create and activate a new Python Virtual Environment.
 1. Clone this repository and `cd` into it.
-2. Run `pip install -r requirements.txt` to install the requirements specified in `requirements.txt`. If using python 3, use `requirements3.txt` instead.
+2. Run `pip install -r requirements.txt` to install the requirements specified in `requirements.txt`.
```
jolyonb, line 42: I think this needs updating now.

README.md
```diff
@@ -6,7 +6,7 @@ A library of graders for edX Custom Response problems.
 
 Version 2.4.1 ([changelog](docs/changelog.md))
 
-Copyright 2017-2022 Jolyon Bloomfield and Chris Chudzicki
+Copyright 2024 Jolyon Bloomfield and Chris Chudzicki
```
jolyonb, line 9: Think this should be 2017-2024

jolyonb (changes requested):
Very minor comments.

Also, there's a line in `docs/index.md` that references that the code works in python 2.

docs/faq.md
```diff
@@ -6,4 +6,4 @@ Unfortunately, no. Those problem types cannot be used in a `customresponse` prob
 
 - Does the grader work with python 2 or python 3?
```
MaferMazu, line 7: I edited this FAQ, but I don't know if it is still required to have that question.

What do you think @jolyonb?

jolyonb: Heh, yeah, can probably remove that now lol!

jolyonb:
I always did a separate PR where I bumped version numbers, updated change logs, zip files, created new tags, etc.

jolyonb (approved):
I'm just going to approve this; I suspect there'll be other minor updates to docs in the future, but this is a fine place for the moment.

PR #329, opened 2024-11-15
chore: drop six and refactor

MaferMazu:
### Screenshots (if appropriate):

- [ ] Desktop screenshots
- [ ] Mobile width screenshots


mitxgraders/plugins/defaults_sample.py
```diff
@@ -9,18 +9,18 @@
...
+# # Here are all of the graders that you can specify defaults for
+# from mitxgraders.stringgrader import StringGrader
+# from mitxgraders.baseclasses import AbstractGrader, ItemGrader
+# from mitxgraders.listgrader import ListGrader, SingleListGrader
+# from mitxgraders.formulagrader.integralgrader import IntegralGrader
+# from mitxgraders.formulagrader.formulagrader import FormulaGrader, NumericalGrader
+# from mitxgraders.formulagrader.matrixgrader import MatrixGrader
+
+# # These will be needed to set attempt-based credit course-wide
+# from mitxgraders.attemptcredit import LinearCredit, GeometricCredit, ReciprocalCredit
```
BryanttV, line 23: Why are we commenting on this?

mitxgraders/helpers/calc/robust_pow.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import print_function, division, absolute_import, unicode_literals
+
 
 import numpy as np
```
BryanttV, line 3:
```suggestion
```

BryanttV: If we want to remove all unused imports we could use [ruff](https://docs.astral.sh/ruff/). For example, in the root folder run: [`ruff check --select F401`](https://docs.astral.sh/ruff/rules/unused-import/). This command shows all unused imports in the code. Just a suggestion.

MaferMazu: Thanks for the suggestion. I used it and found imports in the `__init__` and the `np` in the `robust_pow` file, for which you suggested the change. 

But from the imports in `__init__ ` are needed for the `python_lib.zip`, and in the `robust_pow` the `np` is needed because in the test we are running the docs here https://github.com/eduNEXT/mitx-grading-library/blob/mfmz/drop-six/mitxgraders/helpers/calc/robust_pow.py#L23.

mitxgraders/plugins/defaults_sample.py
BryanttV: I see. Wouldn't it be better to leave all this in the docstring?

mitxgraders/helpers/calc/robust_pow.py
BryanttV: I understand that the test is being run using `doctest`? If so, now I know why we must import it.

mitxgraders/plugins/defaults_sample.py
MaferMazu: I think letting that file without changes is better because I commented the imports to avoid unused imports. Still, for the instructions in the docstring, the idea is to duplicate that file and use it to override things, and if I comment on the imports, we need to change the instructions, and the person should uncomment the things by themselves.

I'll revert this change.

BryanttV (approved):
LGTM

blarghmatey (approved):

PR #331, opened 2024-12-26
feat: add integration tests

MaferMazu:
### Description (What does it do?)
This PR adds a workflow that validates functionality based on the named releases.

Based on the [library documentation](https://mitodl.github.io/mitx-grading-library/):

> The mitxgraders Python library provides a number of configurable Python classes that can be used as graders in edX custom response problems.

So, this PR launches a Tutor env with codejail and runs a script containing the Python classes used by this library.

This PR only tests that you can use the principal classes, but it doesn't test the correctness of each class; for that, we have the unit tests.

> [!IMPORTANT]
> Adding this workflow increases the execution time.
> The unit test runs in seconds, and this integration test runs in ~9 minutes.

> [!NOTE]  
> This test shows that the `python_lin.zip` file works as expected in a Tutor environment. If the `python_lin.zip` doesn't change in a PR, the changes in that PR won't be tested.

#### Detailed changes:
- Added a GitHub action that launches a Tutor env with codejail and imports the MIT course. (The MIT course uses the python_lib.zip file from the branch of a PR)

### How can this be tested?
- Review the checks run.
- Review checks when the test fails: #332

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -34,13 +34,6 @@ jobs:
       - name: Install Tutor
         run: pip install "tutor${{ matrix.tutor_version }}"
         shell: bash
-      
-      - name: Install, enable and initialize Tutor Codejail Plugin
-        run: |
-          pip install git+https://github.com/edunext/tutor-contrib-codejail
-          tutor plugins enable codejail
-          tutor local do init --limit codejail
-        shell: bash
 
       - name: Mount Integration Test
         run: tutor mounts add cms:mitx-grading-library/integration_tests/integration_test.py:/openedx/edx-platform/integration_test.py
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -1,6 +1,6 @@
 name: Integration Test
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   edx-platform-integration-test:
@@ -43,8 +43,7 @@ jobs:
 
       - name: Import MITx Demo Course
         run: |
-          REPO_URL="https://github.com/${{ github.repository }}"
-          tutor local do importdemocourse -r $REPO_URL -d course -v ${{ github.ref_name }}
+          tutor local do importdemocourse -r ${{ github.event.pull_request.head.repo.clone_url }} -d course -v ${{ github.event.pull_request.head.ref }}
         shell: bash
 
       - name: Run integration tests
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -1,0 +1,53 @@
+name: Integration Test
+
+on: [push, pull_request]
+
+jobs:
+  edx-platform-integration-test:
+    name: Integration with Tutor
+    strategy:
+      matrix:
+        # Open edX Version: Sumac
+        tutor_version: ["<20.0.0"]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          path: mitx-grading-library
+
+      - name: Adjust permissions to execute Tutor commands
+        run: |
+          chmod 777 . -R
+        shell: bash
+
+      - name: Set Tutor environment variables
+        run: |
+          cat <<EOF >> "$GITHUB_ENV"
+          LMS_HOST=local.edly.io
+          CMS_HOST=studio.local.edly.io
+          TUTOR_ROOT=$(pwd)
+          EOF
+        shell: bash
+
+      - name: Install and prepare Tutor, Codejail and tests
+        run: |
+          pip install "tutor${{ matrix.tutor_version }}"
+          pip install git+https://github.com/edunext/tutor-contrib-codejail
+          tutor config save
+          tutor plugins enable codejail
+          tutor local do init --limit codejail
+          tutor mounts add cms:mitx-grading-library/integration_tests/integration_test.py:/openedx/edx-platform/integration_test.py
+          tutor local launch -I
+        shell: bash
+
+      - name: Import MITx Demo Course
+        run: |
+          REPO_URL="https://github.com/${{ github.repository }}"
+          tutor local do importdemocourse -r $REPO_URL -d course -v ${{ github.ref_name }}
+        shell: bash
+
+      - name: Run integration tests
+        run: |
+          tutor local run cms python3 integration_test.py
+        shell: bash
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -1,6 +1,9 @@
 name: Integration Test
 
-on: [pull_request]
+on:
+  schedule:
+    # Montlhy at 00:00
+    - cron: '0 0 1 * *'
 
 jobs:
   edx-platform-integration-test:
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -34,6 +34,13 @@ jobs:
       - name: Install Tutor
         run: pip install "tutor${{ matrix.tutor_version }}"
         shell: bash
+      
+      - name: Install, enable and initialize Tutor Codejail Plugin
+        run: |
+          pip install git+https://github.com/edunext/tutor-contrib-codejail
+          tutor plugins enable codejail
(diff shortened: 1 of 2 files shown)


MaferMazu:
If the execution time is not a problem, we can maintain this as it is, but if not, we can run the integration test only if the PR has a specific label. What do you think @blarghmatey?

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Install and prepare Tutor, Codejail and tests
+        run: |
+          pip install "tutor${{ matrix.tutor_version }}"
+          pip install git+https://github.com/edunext/tutor-contrib-codejail
+          tutor config save
+          tutor plugins enable codejail
+          tutor local do init --limit codejail
+          tutor mounts add cms:mitx-grading-library/integration_tests/integration_test.py:/openedx/edx-platform/integration_test.py
+          tutor local launch -I
+        shell: bash
```
mariajgrimaldi, line 42: For a more straightforward debugging process, consider splitting this step into multiple ones depending on their responsibility. What do you think?

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,52 @@
...
+          EOF
+        shell: bash
+
+      - name: Install and prepare Tutor, Codejail and tests
+        run: |
+          pip install "tutor${{ matrix.tutor_version }}"
+          pip install git+https://github.com/edunext/tutor-contrib-codejail
+          tutor config save
+          tutor plugins enable codejail
+          tutor local do init --limit codejail
```
mariajgrimaldi, line 39: Do we need to initialize codejail before launching?

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,52 @@
...
+        shell: bash
+
+      - name: Import MITx Demo Course
+        run: |
+          tutor local do importdemocourse -r ${{ github.event.pull_request.head.repo.clone_url }} -d course -v ${{ github.event.pull_request.head.ref }}
+        shell: bash
+
+      - name: Run integration tests
+        run: |
+          tutor local run cms python3 integration_test.py
```
mariajgrimaldi, line 51: `python` is an alias to `python3` so we could use `python` instead:
```suggestion
          tutor local run cms python integration_test.py
```

integration_tests/integration_test.py
```diff
@@ -0,0 +1,131 @@
...
+        python_path=python_path,
+        extra_files=extra_files,
+        slug="integration-test",
+        limit_overrides_context=course_key_str,
+        unsafely=False,
+    )
+
+
+if __name__ == "__main__":
+    course_key_str = "course-v1:MITx+grading-library+course"
```
mariajgrimaldi, line 130: Could this be a variable provided by the environment?

integration_tests/integration_test.py
```diff
@@ -0,0 +1,131 @@
...
+        all_code (str): The code to be executed.
+        course_key_str (str): The string representation of the course key.
+
+    Returns:
+        None
+    """
+    course_key = CourseKey.from_string(course_key_str)
+    sandbox_service = SandboxService(
+        course_id=course_key,
+        contentstore=contentstore)
```
mariajgrimaldi, line 108:
```suggestion
    sandbox_service = SandboxService(
        course_id=course_key,
        contentstore=contentstore
    )
```

integration_tests/integration_test.py
```diff
@@ -0,0 +1,131 @@
...
+from opaque_keys.edx.keys import CourseKey
+
+# Set up Django environment
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "cms.envs.test")
+django.setup()
+
+log = logging.getLogger(__name__)
+
+# Define the code to be executed
+all_code = """
```
mariajgrimaldi, line 22:
```suggestion
GRADING_CLASSES_CODE = """
```

integration_tests/integration_test.py
```diff
@@ -0,0 +1,131 @@
...
+        'lower': 2,
+        'upper': 3
+    },
+    variables=['a', 'b']
+)
+
+"""
+
+
+def execute_code(all_code, course_key_str):
```
mariajgrimaldi, line 94: We could drop this argument if all_code is a constant:
 ```suggestion
def execute_code(course_key_str):
```

File: integration_tests/integration_test.py (hunk 6 above)
Comment (mariajgrimaldi) [reply integration_tests/integration_test.py:22]: What would be the easiest way to maintain this?

**Review by mariajgrimaldi — CHANGES_REQUESTED** [review]

File: .github/workflows/integration-test.yaml (hunk 1 above)
Comment (MaferMazu) [reply .github/workflows/integration-test.yaml:42]: Thanks for this.

File: .github/workflows/integration-test.yaml (hunk 2 above)
Comment (MaferMazu) [reply .github/workflows/integration-test.yaml:39]: If we want to use codejail, we should init it before launch, following these instructions: https://github.com/eduNEXT/tutor-contrib-codejail?tab=readme-ov-file#usage. If not, the codejail container will not find the app armor profile.

About why we need codejail. It is because the MITx course uses the `mitx-grading-library` `python_lib.zip` file to extend the functionality and uses codejail to run the problem in a safe environment. Principally, because without codejail, the unit fails.

image

File: integration_tests/integration_test.py (hunk 6 above)
Comment (MaferMazu) [reply integration_tests/integration_test.py]: We can use something like the following:
```python
from mitxgraders import *

StringGrader()
FormulaGrader()
NumericalGrader()
MatrixGrader()
.
.
.
```
or
```python
from mitxgraders import (
    StringGrader,
    FormulaGrader,
    NumericalGrader,
    MatrixGrader
)
```
That could test that we can use the Grader functions we want.

I added more context to the code to cover a little more in the tests, using some examples from the unit tests. However, this may not be necessary since it has already been tested in the unit tests. We can test only the fact that we can use the graders. I think this could be totally up to @blarghmatey.

@blarghmatey, do you prefer a code with some examples but with more maintenance or only a test that we can use the graders, which is the main functionality, with less maintenance,  and leave the rest to the unit test?

Thanks, @mariajgrimaldi, for that observation.

.github/workflows/integration-test.yaml
MaferMazu: Only to test it, I tried to run it without codejail, and the result was:
image

It runs the code using this function: https://github.com/openedx/codejail/blob/master/codejail/safe_exec.py#L240.

The test doesn't break, but as we expect a safe environment to run the Loncapa problems, I will leave the codejail part because it is closer to the real environment.

integration_tests/integration_test.py
MaferMazu: I have been thinking about this, and the best approach is to test with the import and leave the details for the unit tests.

jolyonb: I'd suggest running just a MatrixGrader. That's the class that uses the most external dependencies (scipy, numpy). I agree that so long as those are working, the unit tests can handle the rest.

blarghmatey: I think that makes sense. The main purpose for the integration test is to make sure that we can be aware of any regressions due to underlying dependency changes.

blarghmatey:
In terms of execution time, it's not a major problem if it's slow. That being said, we can also just run it on a ~monthly cadence rather than for every pull request.

integration_tests/integration_test.py
MaferMazu: @blarghmatey, for dependency changes, the unit test we are running in python3.8 y 3.11 should be enough because we are using in the test the `edx-sandbox` requirements that codejail uses for redwood and sumac, respectively.

MaferMazu:
@blarghmatey @jolyonb @mariajgrimaldi this is ready for a re-review. ✨

mariajgrimaldi (approved):
Thank you for addressing all my comments!

MaferMazu:
@blarghmatey @jolyonb @pdpinch, I just wanted to follow up on this.

Can you review this PR?

MaferMazu:
Thanks for your feedback. I'll merge this, considering that the comments were addressed and we received feedback in Slack that everything looks good.

@blarghmatey @jolyonb @pdpinch If there is something left, let me know.

Note: this PR doesn't need a new release because it only creates a GitHub workflow.

